PR #5, opened 2025-05-22
[EFA] Enable per-GPU proxy and shared-GPU proxy

MaoZiming:


YangZhou1997:
Mark on it, I will look at it tomorrow.

efa/transport.cc
```diff
@@ -1643,6 +1643,196 @@ std::string UcclEngine::status_to_string(bool abbrev) {
     return s;
 }
 
+bool Endpoint::initialize_engine_by_gpu_idx(int gpu_idx) {
```
YangZhou1997, line 1708: The initialize_engine_by_gpu_idx() function implementation looks pretty similar to the Endpoint() construct function. Maybe extract common functionality into a third function and call that function inside both initialize_engine_by_gpu_idx() and Endpoint()? so later we just need to maintain the common functionality in one place.

YangZhou1997 (commented):
Just a quick comment on code refactoring. The others LGTM.

MaoZiming:
Refactored. PTAL

YangZhou1997:
LGTM! Please directly merge.

PR #15, opened 2025-06-03
(WIP) Adding test script for using CPU proxy to initiate GPUDirect Communication

MaoZiming:


MaoZiming:
Will merge first. Working on multiple CPU proxies in another PR (in the same directory)

PR #30, opened 2025-06-07
Test and evaluate on 3 HGX nodes over 400G IB

YangZhou1997:


YangZhou1997:
uccl (rdma_cuda) runs smoothly on 2 IB nodes (nvlink on): 
image

It is slightly better than nccl: 
image

But uccl fails running on 3 IB nodes, getting hung.

YangZhou1997:
Well, uccl on 3 IB nodes sometimes works: 
image

Not bad when compare to nccl: 
image

rdma_cuda/transport_config.h
```diff
@@ -10,15 +10,13 @@
 
 /// Interface configuration.
 static char const* IB_DEVICE_NAME_PREFIX = "mlx5_";
-static constexpr bool ROCE_NET = true;
+static constexpr bool ROCE_NET = false;
```
MaoZiming, line 13: Maybe in a separate PR we can do something like to avoid changing `ROCE_NET` across setups. 
```
    int is_roce = 0;

    struct ibv_port_attr port_attr;
    if (ibv_query_port(context, 1, &port_attr)) {
        perror("Failed to query port");
        exit(1);
    }

    if (port_attr.link_layer == IBV_LINK_LAYER_ETHERNET) {
        printf("RoCE detected (Ethernet)\n");
        is_roce = 1;
    } else if (port_attr.link_layer == IBV_LINK_LAYER_INFINIBAND) {
        printf("InfiniBand detected\n");
        is_roce = 0;
    } else {
        printf("Unknown link layer: %d\n", port_attr.link_layer);
        exit(1);
    }
```

YangZhou1997: Agree. We need to automate all parameters for IB and RoCE.

YangZhou1997:
The hang bug on 3 IB seems also fixed by https://github.com/uccl-project/uccl/pull/67. Our performance matches the NCCL performance. CC'ing @zhongjiechen @MaoZiming 

UCCL: image
NCCL: image

Testing commands: 
```
bash  ./run_ib.sh 1 24 1 1 8 hosts_ib_multi_process
bash  ./run_ib.sh 1 3 8 1
```

zhongjiechen:
The slightly higher small message latency is because I temporarily disabled bypassing multipath for small messages. It also seems to be the main cause of the hang bug. I will bring it back when I rework #58 .

PR #33, opened 2025-06-09
P2P Transfer based on GPUDirect and CPU Proxy

MaoZiming:
## Run (two nodes)
```bash
./benchmark_remote 0 192.168.0.100 # sender
./benchmark_remote 1 192.168.0.58 # receiver
```

## Results
```
Per-block avg latency:
  Block 0 : 8.090 µs over 20000 ops
  Block 1 : 8.325 µs over 20000 ops
  Block 2 : 7.745 µs over 20000 ops
  Block 3 : 8.163 µs over 20000 ops
  Block 4 : 7.700 µs over 20000 ops
  Block 5 : 7.904 µs over 20000 ops
  Block 6 : 8.156 µs over 20000 ops
  Block 7 : 7.964 µs over 20000 ops

Overall avg GPU-measured latency  : 8.006 µs
Total cycles                       : 2536240765
Total ops                          : 160000
End-to-end Wall-clock time        : 63.720 ms
Throughput                        : 2.51 Mops/s
```


p2p/include/common.hpp
```diff
@@ -0,0 +1,61 @@
...
+      exit(1);                                                \
+    }                                                         \
+  } while (0)
+
+#define MEASURE_PER_OP_LATENCY
+#define kQueueSize 16
+#define kQueueMask (kQueueSize - 1)
+#define kBatchSize 4
+#define kIterations 20000
+#define kNumThBlocks 16
```
YangZhou1997: Although RDMA QPs are thread-safe, using them across threads will incur high thread synchronization overhead (that's probably why you see 30+ microseconds per-op latency). I have tested that if you set kNumThBlocks to 1, you will get ~5 microsecond latency. So ideally, we should have one QP for one CPU thread.

p2p/src/rdma.cpp
```diff
@@ -0,0 +1,433 @@
...
+  int ne = ibv_poll_cq(cq, kMaxOutstandingSends, wc);
+  for (int i = 0; i < ne; ++i) {
+    if (wc[i].status != IBV_WC_SUCCESS) {
+      fprintf(stderr, "CQE error wr_id=%llu status=%s\n",
+              (unsigned long long)wc[i].wr_id, ibv_wc_status_str(wc[i].status));
+      std::abort();
+    }
+  }
+  g_completed.fetch_add(ne, std::memory_order_relaxed);
+  printf(
```
YangZhou1997: Remember to comment out this printf for benchmarking.

p2p/include/common.hpp
MaoZiming: Added!

MaoZiming:
@YangZhou1997 Merging a version first. Tune the parameter next.

PR #40, opened 2025-06-11
Refactor `node_ips` into a folder. 

MaoZiming:


YangZhou1997:
That looks great! Please also change the `python rsync.py` command that appears in various readmes---I think right now we need to specify -n, or keeping the  default is fine?

MaoZiming:
It should support the default in `scripts/node_ips/default.txt` (but users need to fill in the ips of their cluster, which is needed either way)

PR #42, opened 2025-06-11
Utils Refactoring

MaoZiming:
Refactored identical utils files into `include/util/`
I am able to compile afxdp, rdma, and efa.


MaoZiming:
@YangZhou1997 There are some terminology in efa and rdma folders that are i think almost identical, e.g. `flow_map` and `kPortEntropy` in `rdma/util_timer.h`, and `qpw_map_` and `kMaxPath` in `util_timer.h`,

MaoZiming:
I only merged identical header files, and avoided files that look largely similar (but have differently named variables, like above). These should be merged later to avoid confusion for contributors.

include/util/util.h
```diff
@@ -230,18 +232,35 @@ class UINT_CSN {
...
-  std::atomic<bool> done;   // Sync cv wake-up.
-  uint64_t timestamp;       // Timestamp for request issuing.
-  PollCtx() : fence(false), done(false), timestamp(0){};
+  std::atomic<bool> fence;               // Sync rx/tx memcpy visibility.
+  std::atomic<bool> done;                // Sync cv wake-up.
+  std::atomic<uint16_t> num_unfinished;  // Number of unfinished requests.
+  uint64_t timestamp;                    // Timestamp for request issuing.
+  uint32_t engine_idx;                   // Engine index for request issuing.
+#ifdef POLLCTX_DEBUG
+  FlowID flow_id;   // Flow ID for request issuing.
```
YangZhou1997: This FlowID will not compile if POLLCTX_DEBUG is on. But I agree we could extract this common utility out, so let's delete the flow_id and req_id fields. Also, try searching around POLLCTX_DEBUG to delete the code there

include/util/util.h
```diff
@@ -1,6 +1,8 @@
 #pragma once
-
-#include "util_jring.h"
+#ifdef USE_EFA
```
YangZhou1997: Probably call it USE_CUDA?

format.sh
```diff
@@ -4,7 +4,7 @@
 set -e
 
 # Directories to format (excluding thirdparty/, scripts/, doc/, etc.)
-DIRECTORIES=("afxdp" "efa" "gpu_driven" "rdma" "misc")
+DIRECTORIES=("afxdp" "efa" "gpu_driven" "rdma" "misc" "include")
```
YangZhou1997: Should be `DIRECTORIES=("afxdp" "efa" "gpu_driven" "rdma" "misc" "p2p" "include")`

YangZhou1997 (commented):
Thank you @MaoZiming so much for taking the effort to refactor the code. This makes the structure much cleaner!

include/util/util.h
MaoZiming: This currently only seems to be used in EFA folder, will add that flag to rdma too

include/util/util.h
MaoZiming: Okay will delete the code related to `POLLCTX_DEBUG`, these are mostly under EFA code.

PR #47, opened 2025-06-12
Adding write with immediate for remote proxy nvlink forwarding

MaoZiming:


MaoZiming:
https://github.com/uccl-project/uccl/blob/p2p-remote-nvlink/p2p/src/rdma.cpp#L242
@YangZhou1997 For some reason, if I `ibv_open_device` based on NUMA ( e.g. commenting out this line) it will lead to worse performance.

YangZhou1997:
There are several `while` loops in the `cpu_proxy()` function that could delay RDMA polling and increase latency. 

For example, `while (load_volatile_u64(&rb->head) == my_tail){...}` will block the CPU actions of setting done signals to GPU. Another code `while (cmd == 0)` will also likely delay RDMA operations or CPU actions. What we can do is first doing an `if` check to see if there is any work from GPU; if not, we should check RDMA cq once; if no cq, we then check GPU work again. 

Basically, one principle of getting low latency is to avoid any blocking CPU code. This way, we should be able to get much lower latency. Some examples might be: 

https://github.com/uccl-project/uccl/blob/287bf4f0e977bb4e1fdd08eb9941d7e6530e623a/rdma/transport.cc#L359-L382

MaoZiming:
@YangZhou1997 I pushed a new commit and turned `while (load_volatile_u64(&rb->head) == my_tail)` into `if`. Also cleaned the code with `while(cmd==0)` (https://github.com/uccl-project/uccl/pull/47/commits/2b9d6dfca48cdd0114f7537c73915f4c0a59df0d#diff-d7796e0eacf612f82ce8fb1011072a53ff7181ebb78a24fc618412f654354650L170). the latency and throughput is still roughly the same.

p2p/src/proxy.cpp
```diff
@@ -57,7 +127,7 @@ void cpu_proxy(RingBuffer* rb, int block_idx, void* gpu_buffer,
   for (size_t seen = 0; my_tail < kIterations;) {
     poll_completions(cq, finished_wrs, finished_wrs_mutex);
```
YangZhou1997, line 267: Here needs to immediately signal the GPU if there are any finished WRs.

p2p/src/proxy.cpp
```diff
@@ -67,7 +137,8 @@ void cpu_proxy(RingBuffer* rb, int block_idx, void* gpu_buffer,
       }
 #endif
       /* spin */
-      _mm_pause();
+      // _mm_pause();
+      continue;
```
YangZhou1997: It seems this will skip some requests?

YangZhou1997: Oh, I think it won't

MaoZiming:
@YangZhou1997 Refactor-ed code based on the comments. still
```
Overall avg GPU-measured latency  : 29.650 µs
Total cycles                       : 920532954203
Total ops                          : 15839999
End-to-end Wall-clock time        : 7136.878 ms
Throughput                        : 2.24 Mops
```
I am also looking into how to optimize this further

YangZhou1997:
Thank you Ziming for testing them out! I think we can start from reducing the latency of 1 SM as batch size increases. Now, 4 batch size of 1 SM has 19us, while 1 batch size only has 12us. This increase seems too high for me. I will also look into it!

YangZhou1997:
I saw you added kHeadTailLimit, which is like kMaxInflight that allows constant inflight requests (which is exactly what we want for tput measurement). Nice code! 

In branch [p2p-yang](https://github.com/uccl-project/uccl/tree/p2p-yang), I added `IBV_ACCESS_RELAXED_ORDERING` when registering memory so as to avoid the performance anomaly you see when choosing NICs. This also helps reach 400G.   There are a few other changes, such as calculating tput using GPU time, and you can check them out. 

I have measured: under 1 SM and 32 max inflight: 
| Batch size | Per-op latency (us) | RDMA RTT (us) | Op rate (Mops) |
| --- | --- | --- | --- |
| 1 | 11.9 | 5.9 | 0.14 |
| 2 | 16.9 | 6.5 | 0.28 |
| 4 | 17.2 | 7.0 | 0.49 |
| 8 | 22.7 | 8.0 | 0.95 |
| 16 | 26.2 | 10.0 | 1.61 |
| 32 | 26.6 | 10.2 | 1.58 |

So it seems that a 16 batch size is a good spot to balance tput and per-op latency. 

Under 16 batch size, I try to increase the number of SMs: 
| Num of SMs | Per-op latency (us) | RDMA RTT (us) | Op rate (Mops) |
| --- | --- | --- | --- |
| 1 | 26.2 | 10.0 | 1.61 |
| 2 | 25.9 | 11.4 | 3.13 |
| 4 | 27.9 | 13.5 | 5.41 |
| 8 | 49.3 | 33.0 | 5.77 |
| 16 | 91.5 | 75.8 | 5.95 |

So it seems that 4 CPU cores is a good spot for balance tput and CPU usage. 

Overall, CPU proxy adds around 6- 16us latency, which is a pretty good result!

MaoZiming:
@YangZhou1997 Thank you! 
Interesting, `IBV_ACCESS_RELAXED_ORDERING` has an effect on which NIC to pick. Do you happen to know why?
Also, `IBV_ACCESS_RELAXED_ORDERING` + picking the right NIC have a huge impact on performance, for example, 4 CPUs + batch size = 16 without these two --> 
```
Overall avg GPU-measured latency  : 62.361 µs
Total cycles                      : 488959407460
Total ops                         : 3959990
End-to-end wall-clock time        : 1752.615 ms
Throughput                        : 2.26 Mops
```

MaoZiming:
Num of SMs | Per-op latency (us) | RDMA RTT (us) | Op rate (Mops)
-- | -- | -- | --
4 | 27.9 | 13.5 | 5.41

Still, here the WR completion time is `13.21 us`, maybe we can still reduce that

YangZhou1997:
Yes, we discovered the `IBV_ACCESS_RELAXED_ORDERING`  issue when running uccl on IBM testbed: https://github.com/NVIDIA/nccl/issues/1731. Without relaxed ordering, GPU and NIC under the same pcie switch can only achieve 84Gbps, but under different pcie switches can achieve 200Gbps. With relaxed ordering, under the sane pcie switch can reach 400Gbps, under different pcie swtiches keeps 200Gbps. 

Relaxed ordering reorders pcie transactions so getting higher performance, and is critical for GPUDirect performance for some testbed. (we also find on our original HGX servers, there is no need for relaxed order)

MaoZiming:
@YangZhou1997 Thanks for the pointer! It seems `PIX` still doesn't give the best performance. For example, if I pick `PHB` (`NIC5` for `GPU0`), it leads to even higher throughput. 
I think this is probably fine for now, since 5.5 MOps * 8 KB = 360 Gbps (which is close to the line rate).

PR #51, opened 2025-06-15
Remote CPU Proxy NVLink-based forwarding

MaoZiming:
```
Per-block avg latency:
  Block 0 : 27.358 µs over 989999 ops
  Block 1 : 27.680 µs over 989994 ops
  Block 2 : 27.674 µs over 989999 ops
  Block 3 : 27.213 µs over 989998 ops

Overall avg GPU-measured latency  : 27.481 µs
Total cycles                      : 215473249225
Total ops                         : 3959990
End-to-end wall-clock time        : 671.455 ms
Ops Throughput                    : 5.90 Mops
Total Throughput                  : 386.51 Gbps
```


YangZhou1997:
I just did some benchmarking for cudaMemcpyPeerAsync: 

```
// 1MB msg size:
cudaMemcpyPeerAsync performance: 154.629 GB/s

// 8KB msg size:
cudaMemcpyPeerAsync performance: 2.29951 GB/s
```

Basically, it does not seem possible to catch up with NVLink bandwidth with a small message size like 8KB. We need roughly MB of messages. 

Therefore, here comes a research problem: in DeepEP normal mode, how could we let multiple src GPUs dispatch their messages to a specific remote GPU such that their RDMA dst addresses are contiguous? This way, we could just do a big cudaMemcpyPeerAsync to finish the forwarding. 

In each batch, each GPU has 4096 tokens, each with 7168 hidden size to dispatch, so each GPU should also receive an average of 4096 * 7168 = 28MB of data. If we can distributedly coordinate these src GPUs to send data to other GPUs in a remote-contiguous way, we should be able to get enough message size. Note that we can tolerate some gaps/waste in this NVLink-based copy, as the NVLink bandwidth is huge.

MaoZiming:
Agree. We need coalescing on the remote node for reasonable throughput.

YangZhou1997:
It seems someone changed the nvcc to v11.x that does not support sm-90. So I modified the makefile a bit to use nvcc under CUDA_HOME, this should also address the CUDA toolchain mismatch problems.

YangZhou1997:
Wow, this performance is impressive! How many SM are you using for nvlink forwarding?

YangZhou1997:
Also, looking at the code, it seems that the latency does not include the NVlink forwarding latency. The number of SMs could be up to 16?

MaoZiming:
@YangZhou1997  I am using `kNumThBlocks` = 4 number of threads to do copy, each one will launch a kernel (with `RECEIVER_BATCH_SIZE` = 16 blocks and 256 threads) for copying.  NVLink forwarding is done currently asynchronously so the latency does not include that (in other words, the kernel on the sender does not wait for the forwarding to finish before proceeding).

MaoZiming:
Also I think the next step is to think about how the sender gets notified when the NVLink forwarding finishes.

YangZhou1997:
I see, so in total 4 * 16 = 64 SMs now. I guess using more efficient copy kernel will reduce the SM usage, eg, https://gist.github.com/happierpig/09e5f1387aa84680559763432d3107e1

MaoZiming:
@YangZhou1997 I manage to get it to 4 * 2 = 8 SMs now. I will incorporate the copy kernel tomorrow and do more parameter tuning. Merging first.

PR #60, opened 2025-06-18
Testing IB cluster stuck issue

MaoZiming:


rdma/transport_config.h
```diff
@@ -6,16 +6,16 @@
 #include <unistd.h>
 
 // #define STATS
-// #define LAZY_CREATE_ENGINE
+#define LAZY_CREATE_ENGINE
```
MaoZiming, line 9: Single/multi-process should work. The cluster is crowded so sometimes it gets stuck.
```
bash ./run_nccl_test.sh 1 16 1 1 8 hosts_multi_process
bash ./run_nccl_test.sh 1 2 8 1
```

PR #62, opened 2025-06-19
specify more files for building in afxdp

YangZhou1997:


YangZhou1997:
TPU gvNIC does not support zero-copy AFXDP, therefore it will degrade into SKB mode. 

Also, our current AFXDP-based UCCL transport can only work for one core on TPU VMs. The problem happens when the client and server are negotiating the dst ports for stable RSS mapping, where with two cores, RSS probe rsp packets cannot reach destinations. 

Our AFXDP implementation can achieve ~5Gb/s because of the slow SKB mode. We need to switch to DPDK.

YangZhou1997:
Also, afxdp socket passing via UDS requires a relatively new kernel. You can use the following to upgrade kernel to v6.8.0

```
sudo apt update
sudo apt install linux-image-generic-hwe-22.04
sudo apt install -y linux-headers-$(uname -r) build-essential
```

PR #64, opened 2025-06-21
make it work on AMD + Broadcom NICs

YangZhou1997:


YangZhou1997:
With three nodes (tw051, tw036, tw003) each with 4 GPUs (0,1,2,4), four net_peer, RCCL gives poor performance. CC'ing @zhongjiechen 

nvlink off: image
nvlink on: image

YangZhou1997:
The AMD cluster is a bit messy with four RDMA NICs down, which is why UCCL triggers "port inactive" bugs when initing the RDMA factory. 

Cluster GPU to RDMA dev mapping: 
Dev 0 -> 75:00 -> rdma3 up
Dev 1 -> 05:00 -> rdma0 up
Dev 2 -> 67:00 -> rdma2 up
Dev 3 -> 17:00 -> rdma1 down
Dev 4 -> f5:00 -> rdma7 down
Dev 5 -> 85:00 -> rdma4 up
Dev 6 -> e7:00 -> rdma6 down
Dev 7 -> 97:00 -> rdma5 down
So we should only use: HIP_VISIBLE_DEVICES=0,1,2,5

YangZhou1997:
Now, the problem is that Broadcom NIC does not support some of the features we requested: https://github.com/uccl-project/uccl/blob/990ee6703e926b4664b2cf57f14e074879be12af/rdma/util_rdma.h#L966-L987

Prompting ChatGPT, and it told me Broadcom NIC does not support "does not support ibv_create_cq_ex() or most other libibverbs extended verbs such as IBV_WC_EX_WITH_SRC_QP". But it does support shared RQ and CQ. 

One can test by `LD_LIBRARY_PATH="${CONDA_LIB_HOME}:/opt/rocm-6.3.1/lib:${LD_LIBRARY_PATH}" ./transport_test --server=true`

YangZhou1997:
Wow, this (running UCCL on Broadcom) is great! @zhongjiechen 

Btw, if you want to test RCCL/UCCL, and you have set `LD_LIBRARY_PATH="${CONDA_LIB_HOME}:/opt/rocm-6.3.1/lib:${LD_LIBRARY_PATH}"`, then you need to `unset LD_LIBRARY_PATH` to avoid conda lib impacting mpi

YangZhou1997:
It also seems that different nodes have different active IB interfaces, by checking `ibstat`...

YangZhou1997:
One >=2 GPUs per node, UCCL would hit `transport.cc:473] Check failed: rdma_ctx_map_.find(ctrl_work.peer_id) == rdma_ctx_map_.end()`

Issue #66, opened 2025-06-21
rdma transport_test hang on uccl_connect/accept


YangZhou1997:
Hi @zhongjiechen, 

I am developing a KV transport engine based on the transport.h/cc for Ray. However, I find transport_test in the current commit hangs on uccl_connect/accept, specifically on these LoC: https://github.com/uccl-project/uccl/blob/e361c4e0cf29c1d214cc7c0372be22d1114291f2/rdma/transport.cc#L998-L1012 https://github.com/uccl-project/uccl/blob/e361c4e0cf29c1d214cc7c0372be22d1114291f2/rdma/transport.cc#L1103-L1118

Any thoughts on how to fix this? 

Best,
Yang

YangZhou1997:
I also tested that It works on HEAD^2

zhongjiechen:
Hi @YangZhou1997 . This is because i made an assumption about the way that application makes connections: This works for NCCL, but not for general communication. Let me create a PR to make it work for all cases.

YangZhou1997:
Got you! Thank you, Zhongjie!

zhongjiechen:
Hi @YangZhou1997 , #67 has fixed this BUG. I tested alltoall, allreduce, send/recv (which I think is very similar to transport_test), and it all works.

YangZhou1997:
Yes, it works on my side!

PR #67, opened 2025-06-22
Fix connection setup bug.

zhongjiechen:
This PR fixes the connection setup bug in `rdma`. Currently, we use the following algorithm to establish RDMA context:

1. Each pair of connect/accept will determine the leader based on (IP address, dev index, gpu index).
2. If it is a follower, whether to install the rdma context is completely notified by the leader.
3. If it is the leader, it must be the first call in all connect/accept to install the rdma context.

This PR has successfully run alltoall, allreduce, and sendrecv on multiple nodes.


zhongjiechen:
This PR solves #66 .

PR #76, opened 2025-06-24
Introduce UCCL_PARAM

zhongjiechen:
This PR aims to make UCCL parameter configuration more flexible. `UCCL_PARAM` is introduced to help with this, whose codebase is from NCCL. 

When configuring variables as `UCCL_PARAM`, the following principles should be followed:

1) Variables on the critical path cannot use `UCCL_PARAM` directly (it may affect performance). The appropriate approach is to read `UCCL_PARAM` and assign it to local variables.

2) Pay attention to the necessity of changing constants to `UCCL_PARAM`. For example, `kMaxSRQ`, we generally do not change it, so there is no need to change it to `UCCL_PARAM`. For `kChunkSize`, we need to change it to `UCCL_PARAM`, but we need to pay attention to 1) because it is on the critical path.


zhongjiechen:
We can complete this PR step by step, and it is recommended that the subsequent commits of this PR include the modified variable name.

PR #84, opened 2025-06-24
enlarging chunksize to 512KB but does not help improve perf on amd; f…

YangZhou1997:
…ixing size overflow bugs


YangZhou1997:
To run on AMD+Broadcom platform, follow: https://github.com/uccl-project/uccl/tree/main/rdma#if-your-are-using-broadcom-nics

Basically, 
```
make -f Makefile_hip -j "CXXFLAGS=-DBROADCOM_NIC"
./run_rccl_test.sh
```

YangZhou1997:
I think I have found the root cause of why UCCL performs worse than RCCL. RCCL will actually pair a GPU with a NIC that is not under the same PCIe bridge (checked via `lstopo`). 

For example, from `lstopo`, we have Hip GPU 1 -> 05:00 -> rdma0 up. But RCCL will pair GPU1 with rdma2 (checked via `-x NCCL_DEBUG=INFO`), which gives better performance than rdma0. When I enforce the same pairing to UCCL, UCCL performs similarly or even better than RCCL (four nodes each with one GPU). So we need to figure out why RCCL would pair like this and how to let our plugin support it. 

UCCL: 
image
NCCL: 
image

YangZhou1997:
Fixed by using AMD-specific ibgdr detection. 
image

YangZhou1997:
On two nodes, each with 4 GPUs, for alltoall nvlink off (for stress-testing the network), UCCL beats RCCL by up to 39% (at 32MB message size): 

UCCL: 
image

NCCL: 
image

zhongjiechen:
Great results and huge efforts on debugging! @YangZhou1997

PR #127, opened 2025-07-02
Running uccl on gh200 with pip

MaoZiming:


MaoZiming:
@YangZhou1997 it seems i can't build the rocm version on arm with docker, will need to mod the script

MaoZiming:
https://test.pypi.org/manage/project/uccl/release/0.0.1.post7/
The arm image is pushed to

MaoZiming:
Will need to run `pip install -i https://test.pypi.org/simple/ uccl -U` on all nodes.

YangZhou1997:
Hi Ziming, do we need a new name for the .so of the gh, as it seems to be conflicting with the nccl-cuda one?

YangZhou1997:
Oh, actually, no need. It will have a different arch name in the wheel

PR #139, opened 2025-07-03
Making NCCL more scalable

YangZhou1997:


YangZhou1997:
NCCL nvlink communication (`p2p.cc`) does not go through the CPU proxy. Only net and shm go through the CPU proxy.

YangZhou1997:
It has been implemented in: 

https://github.com/derekwin/rccl/commit/c6b5ae3d73778c57447768270ec9c936abf8ca0c
https://github.com/derekwin/nccl/commit/1ee212199f28217e7133030736379ce999119be7

But not observing too much performance improvement, for example, rccl alltoall across 3*8 GPUs: 

image

PR #144, opened 2025-07-05
Making python binding react to ctrl+c and add async send/recv

YangZhou1997:


YangZhou1997:
Now, uccl_p2p can run (with `--device cpu`) on GH200: 
image

But NCCL is broken: 
image

@MaoZiming , have you encountered this?

YangZhou1997:
Essentially, we do not need async send and recv, which is usually challenging for Python applications to manage polling; what we need is just a vector send and recv API.

MaoZiming:
@YangZhou1997 This works
```
❯ bash ./run_gh200_test.sh uccl 2 1 1
```

YangZhou1997:
> @YangZhou1997 This works
> 
> ```
> ❯ bash ./run_gh200_test.sh uccl 2 1 1
> ```

Oh, I mean the `pip/benchmark_nccl.py`. What is the key setting that makes nccl collectives work on the GH200?

PR #170, opened 2025-07-18
Add benchmark for nixl with mooncake backend

praveingk:
Added an option in benchmark_nixl to choose between UCX and Mooncake backend.
Mooncake doesn't support notifs for metadata exchanges. 
Hence, added an out-of-band zmq socket for control messages.

### DIFF
--- a/p2p/benchmarks/benchmark_nixl.py
+++ b/p2p/benchmarks/benchmark_nixl.py
@@ -5,6 +5,7 @@ import sys
 import time
 from typing import List
 import traceback
+import zmq
 from datetime import datetime
 
 try:
@@ -21,7 +22,7 @@ except ImportError as exc:
 
 
 import numpy as np
-
+listen_port = 9000
 def create_dataset(role, size, device, gpu_idx=0):
     """
     Create a dataset of tensors whose total size is at least size in bytes.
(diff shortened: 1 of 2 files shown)


MaoZiming:
@praveingk Great! Do you observe any performance difference between UCX and Mooncake backend?

praveingk:
@MaoZiming I can get line-rate with mooncake in the 100G nodes. It does throw out some "packet mismatch errors" sometimes since it picks the wrong HCA, mlx5_0 on one node, and mlx5_1 on the other node.
I am yet to test it in a 400G testbed.

p2p/benchmarks/benchmark_nixl.py
```diff
@@ -100,22 +159,70 @@ def initialize_xfer_metadata(
...
+            elif state == "ERR":
+                print("Error in transfer")
+                break
+    else:
+        while True:
+            transfer_done = zmq_socket.recv()
+            if transfer_done.decode("utf-8") == uid:
+                break;
+
 def start_transfer(
```
MaoZiming: Maybe rename this to something like `start_transfer_ucx`

p2p/benchmarks/benchmark_nixl.py
```diff
@@ -201,6 +335,8 @@ def start_agent_pair(size, args):
             register_descs,
         )
         cleanup_agent(agent)
+        if args.backend in "mooncake":
```
MaoZiming:
```suggestion
        if args.backend == "mooncake":
```

YangZhou1997:
Hi @praveingk , is it possible to add a quick instruction on how to install Mooncacke? Eg, something folded under https://github.com/uccl-project/uccl/tree/main/p2p#running-nixl-with-ucx-backend

praveingk:
@YangZhou1997 Yes, I will do that. 👍🏽

PR #179, opened 2025-07-21
Removing the need of specifying ROCE_GID_IDX or IB_GID_IDX

YangZhou1997:


YangZhou1997:
As mentioned in https://github.com/uccl-project/uccl/issues/172

PR #180, opened 2025-07-22
Optimizing nixl benchmark

YangZhou1997:


YangZhou1997:
> Looks good 👍🏽

It seems my cat accidentally hits the request-for-review button 😂. I am still working on it.

YangZhou1997:
NIXL switching from READ to WRITE gives a big improvement on our GH200 servers using host mem over 200G IB: 

NIXL using READ: 
image

NIXL using WRITE will hit some performance drop at 100MB messages (not always, but most time), not sure why: 
image

But still worse than UCCL, especially on medium-sized messages: 
image

YangZhou1997:
I also tried to use one shared NIXL agent with multiple `transfer_handle`s (code in `benchmark_nixl2.py`), but performance is pretty bad, not sure why though: 

image

praveingk:
Interesting, I recently learnt from UCX team that READ has an extra operation, and its slower than WRITE. vllm is currently using READ (from decode->prefix), would be changed to WRITE eventually.

PR #185, opened 2025-07-23
Uccl p2p dual

YangZhou1997:


MaoZiming:
@YangZhou1997 What are some use cases for the UCCL dual send and receive?

YangZhou1997:
Yilong tested and found nccl dual sendrecv suffers from performance issues.
I guess it is because of non zero copy nature and proxy bottleneck in nccl.
So I try to test if uccl p2p would be better. It turns out yes.

PR #186, opened 2025-07-23
Add APIs for NIXL's UCCL backend

praveingk:
The changes are the following to support UCCL backend inside NIXL: 
1) Added `WITH_PYTHON` definitions to provide cpp only APIs.
2) `uccl_engine` APIs which are used by NIXL's UCCL backend for integration (https://github.com/praveingk/nixl-uccl/pull/1). WRITE works with the nixl plugin. READ is still not working (debugging in progress).
3) Added `DISABLE_CALL_ONCE_STATIC` definitions in rdma-transport code to provide clean way to initialize/destroy/re-initialize uccl plugin from another process (nixl).  Additionally, exposing the socket fd used for coordination to be used to additional coordination by the uccl backend for NIXL.
4) `benchmark_nixl.py` to support uccl backend.
4) Bug: Experienced the GPU bus ids were not matching because of Uppercase/Lowercase. Hence, added a normalization.


YangZhou1997:
Great work! Will talk a look very soon!

p2p/engine.cc
```diff
@@ -23,17 +23,21 @@ std::once_flag glog_init_once;
 constexpr uint32_t kGpuStreamId = 0;
 
 inline void check_python_signals() {
+#ifdef WITH_PYTHON
```
YangZhou1997: Hi @praveingk , is this WITH_PYTHON macro required because the NIXL python layer has already handled it?

p2p/engine.cc
```diff
@@ -279,11 +293,11 @@ bool Endpoint::regv(std::vector<void const*> const& data_v,
 }
 
 bool Endpoint::send(uint64_t conn_id, uint64_t mr_id, void const* data,
-                    size_t size, bool inside_python) {
+                    size_t size) {
   DCHECK(size <= 0xffffffff) << "size must be less than 4GB";
-  [[maybe_unused]] auto _ =
```
YangZhou1997, line 310: can you keep this `[[maybe_unused]] auto _ = xxx`, as both python threads or c++ naive thread may call this function? For c++ native threads, `py::gil_scoped_release{}` and `check_python_signals()` is super slow, yet unnecessary.

p2p/engine.cc
```diff
@@ -317,7 +331,7 @@ bool Endpoint::send(uint64_t conn_id, uint64_t mr_id, void const* data,
       done[ureq_issued % kMaxInflightChunks] = false;
       ureq_issued++;
     }
-    auto _ = inside_python ? (check_python_signals(), nullptr) : nullptr;
```
YangZhou1997: Similarly, here. thx

p2p/engine.cc
```diff
@@ -340,11 +354,10 @@ bool Endpoint::send(uint64_t conn_id, uint64_t mr_id, void const* data,
   return true;
 }
 
-bool Endpoint::recv(uint64_t conn_id, uint64_t mr_id, void* data, size_t size,
-                    bool inside_python) {
-  [[maybe_unused]] auto _ =
```
YangZhou1997, line 371: Similarly, here. thx

p2p/engine.cc
```diff
@@ -379,7 +392,7 @@ bool Endpoint::recv(uint64_t conn_id, uint64_t mr_id, void* data, size_t size,
       done[ureq_issued % kMaxInflightChunks] = false;
       ureq_issued++;
     }
-    auto _ = inside_python ? (check_python_signals(), nullptr) : nullptr;
```
YangZhou1997: Similarly, here. thx

p2p/engine.cc
```diff
@@ -643,9 +664,10 @@ bool Endpoint::recvv(uint64_t conn_id, std::vector<uint64_t> mr_id_v,
 }
 
 bool Endpoint::read(uint64_t conn_id, uint64_t mr_id, void* dst, size_t size,
-                    uccl::FifoItem const& slot_item, bool inside_python) {
-  auto _ = inside_python ? (py::gil_scoped_release(), nullptr) : nullptr;
```
YangZhou1997: Similarly, here. thx

p2p/engine.cc
```diff
@@ -663,22 +685,23 @@ bool Endpoint::read(uint64_t conn_id, uint64_t mr_id, void* dst, size_t size,
         static_cast<uccl::UcclFlow*>(conn->uccl_conn_id_.context), mhandle, dst,
         size, slot_item, &ureq);
     if (rc == -1) {
-      auto _ = inside_python ? (check_python_signals(), nullptr) : nullptr;
```
YangZhou1997: Similarly, here. thx

p2p/engine.cc
```diff
@@ -663,22 +685,23 @@ bool Endpoint::read(uint64_t conn_id, uint64_t mr_id, void* dst, size_t size,
...
+      check_python_signals();
       std::this_thread::yield();
     }
   } while (rc == -1);
 
   while (!ep_->uccl_poll_ureq_once(&ureq)) {
-    auto _ = inside_python ? (check_python_signals(), nullptr) : nullptr;
```
YangZhou1997: Similarly, here. thx

p2p/engine.cc
```diff
@@ -727,13 +751,13 @@ void Endpoint::send_proxy_thread_func() {
 
   while (!stop_.load(std::memory_order_acquire)) {
     if (jring_sc_dequeue_bulk(send_task_ring_, &task, 1, nullptr) == 1) {
-      send(task.conn_id, task.mr_id, task.data, task.size, false);
```
YangZhou1997: Similarly, here. thx

p2p/engine.cc
```diff
@@ -727,13 +751,13 @@ void Endpoint::send_proxy_thread_func() {
...
+      send(task.conn_id, task.mr_id, task.data, task.size);
       task.self_ptr->done.store(true, std::memory_order_release);
     }
 
     if (jring_sc_dequeue_bulk(read_task_ring_, &read_task, 1, nullptr) == 1) {
       read(read_task.conn_id, read_task.mr_id, read_task.data, read_task.size,
-           read_task.slot_item, false);
```
YangZhou1997: Similarly, here. thx

p2p/engine.cc
```diff
@@ -745,14 +769,16 @@ void Endpoint::recv_proxy_thread_func() {
 
   while (!stop_.load(std::memory_order_acquire)) {
     if (jring_sc_dequeue_bulk(recv_task_ring_, &task, 1, nullptr) == 1) {
-      recv(task.conn_id, task.mr_id, task.data, task.size, false);
```
YangZhou1997: Similarly, here. thx

p2p/engine.h
```diff
@@ -125,8 +130,7 @@ class Endpoint {
    *   data: the data to send
    *   size: the size of the data
    */
-  bool send(uint64_t conn_id, uint64_t mr_id, void const* data, size_t size,
-            bool inside_python = true);
```
YangZhou1997: Similarly, here. thx

p2p/engine.h
```diff
@@ -138,8 +142,7 @@ class Endpoint {
    *   data: the data to receive
    *   size: the size of the data
    */
-  bool recv(uint64_t conn_id, uint64_t mr_id, void* data, size_t size,
-            bool inside_python = true);
```
YangZhou1997: Similarly, here. thx

p2p/engine.h
```diff
@@ -172,7 +175,7 @@ class Endpoint {
    *   slot_item: the slot item to use for the transfer
    */
   bool read(uint64_t conn_id, uint64_t mr_id, void* dst, size_t size,
-            uccl::FifoItem const& slot_item, bool inside_python = true);
```
YangZhou1997: Similarly, here. thx

rdma/rdma_io.cc
```diff
@@ -32,10 +32,14 @@ int RDMAFactory::init_devs() {
   std::vector<std::tuple<std::string, fs::path, int>> ib_nics_with_dev_idx;
   std::stringstream init_devs_log;
 
+#ifndef DISABLE_CALL_ONCE_STATIC
```
YangZhou1997, line 35: Sorry, if we are doing `std::make_shared<RDMAFactory>()` for every pthread, what would be the use scenario?

rdma/transport.cc
```diff
@@ -687,14 +688,18 @@ void UcclRDMAEngine::handle_install_ctx_on_engine(Channel::CtrlMsg& ctrl_work) {
 RDMAEndpoint::RDMAEndpoint(int num_engines_per_dev)
     : num_engines_per_dev_(num_engines_per_dev),
       stats_thread_([this]() { stats_thread_fn(); }) {
+#ifndef DISABLE_CALL_ONCE_STATIC
```
YangZhou1997, line 691: Similarly here

rdma/transport.cc
```diff
@@ -757,10 +762,12 @@ int try_bind_listen_socket(int* sock_fd, int base_port,
 
 bool RDMAEndpoint::initialize_engine_by_dev(int dev,
                                             bool enable_p2p_listen = false) {
-  static std::vector<std::once_flag> flags_per_dev_(MAX_IB_DEVS);
   bool called = false;
+#ifndef DISABLE_CALL_ONCE_STATIC
```
YangZhou1997, line 766: Similarly here. I am worried that creating engines for each thread would create a huge number of engines in total.

rdma/transport.cc
```diff
@@ -954,9 +962,19 @@ void RDMAEndpoint::install_ctx_on_engines(int fd, int dev, PeerID peer_id,
   DCHECK(factory_dev) << "install_ctx_on_engines: get_factory_dev()";
 
   ret = send_message(fd, &factory_dev->gid.raw, 16);
+  printf("Sent  factory dev raw: %02x:%02x:%02x:%02x:%02x:%02x:%02x:%02x:%02x:%02x:%02x:%02x:%02x:%02x:%02x:%02x\n",
```
YangZhou1997: using `UCCL_LOG_EP`?

rdma/transport.cc
```diff
@@ -1159,7 +1180,26 @@ ConnID RDMAEndpoint::uccl_accept(int dev, int listen_fd, int local_gpuidx,
   auto* factory_dev = RDMAFactory::get_factory_dev(dev);
   DCHECK(factory_dev) << "uccl_accept: get_factory_dev()";
 
+  // Debug: Check if listen_fd is valid
```
YangZhou1997: Wondering if we still need this?

rdma/transport.h
```diff
@@ -1244,8 +1245,14 @@ class UcclFlow {
     memset(&recv_comm_, 0, sizeof(recv_comm_));
     int num_devices = ep->get_num_devices();
     // Avoid all flows using the same initial engine offset.
+
```
YangZhou1997, line 1250: I think this always needs to be static, otherwise, a local `std::vector<std::atomic<uint32_t>>` has no impact on load balancing.

p2p/engine.cc
praveingk: Hi @YangZhou1997 I started with WITH_PYTHON macro for c++ specific consumers although NIXL uses python, and still requires the python signals to be handled. I noticed the bool `Endpoint::send(uint64_t conn_id, uint64_t mr_id, void const* data, size_t size, bool inside_python`, and thought this could be a replacement of that.  But I can remove this macro, and keep the implementation as earlier to keep it simple.

YangZhou1997: I see. I chose `bool inside_python` because we want to provide both sync and async send/recv APIs. When python application uses the sync APIs calling `send`, inside_python must be true to check signal; when python application uses the async APIs, we have a c++ native thread calling `send`, inside_python must be false to avoid check signal (as I find native thread checking python signal is costly). To avoid accommdate the two APIs at the same application, I choose to use dynamic `bool inside_python` instead of static macro. What do you think of? 

Another way is to use c++ template like <bool inside_python>, but it seems hard to work if `send` function is not implemented in the header file.

praveingk: Makes sense @YangZhou1997  Let me revert this file to its previous implementation.

rdma/rdma_io.cc
praveingk: @YangZhou1997  The reason I had to use this is because I was destroying and creating a new nixl agents (1000 times)  dynamically to test (just like mooncake,ucx backends). While doing this, it resulted in some inconsistencies like : new listen_fd (control socket) wasn't getting created for the 2nd time nixl_agent is created, devices were uninitialized during destructor of the 1st nixl_agent, and not getting reinitialized, etc. Please let me know if you have any suggestions on how to deal with such scenario. Let me revisit this again on my part.

For now, I could change `benchmark_nixl` to use just one nixl_agent for the entire duration of the test(which would remove the need to use DISABLE_CALL_ONCE , but my only worry was while mooncake/ucx backends would support dynamic creation of nixl_agents on the go, UCCL backend wouldn't support that, and a process has to use a single nixl_agent for its entire timeline.

YangZhou1997: Oh, I see. Then I think we should keep `DISABLE_CALL_ONCE_STATIC` for now. In the long term, we should fix the reinit issue. Can you open an issue on this (basically copy-paste what you just described) after you merge this PR? Just for tracking purpose~

praveingk:
@YangZhou1997 I have removed the WITH_PYTHON primitive in build, and added a state check `if (PyGILState_Check())` since I am seeing assertion when I call the API from uccl_engine.cc.  Checked its working.
Please let me know if it can be merged.

p2p/Makefile
```diff
@@ -31,36 +34,48 @@ LDFLAGS         = -L$(CUDA_LIB) -lcudart -lcuda $(REDIS_LIBS) \
...
+	$(CXX) $(CAPI_OBJECT) $(CORE_OBJECT) ../rdma/librdma.a \
+	      -L$(CUDA_LIB) -lcudart -lcuda \
+	      -o $@ $(LDFLAGS) $(PYTHON_LDFLAGS) $(CXXFLAGS) \
+	      -Wl,-rpath,$(CUDA_LIB)
 
-../rdma/librdma.a: ../rdma/*.cc ../rdma/*.h
-	make -C ../rdma -j$(nproc)
+
+../rdma/librdma.a: $(filter-out ../rdma/nccl_plugin.cc, $(wildcard ../rdma/*.cc)) ../rdma/*.h
+	make CXXFLAGS="-DDISABLE_CALL_ONCE_STATIC" -C ../rdma librdma.a -j$(nproc)
```
YangZhou1997: I would suggest building two `librdma.a`, one `librdma_for_target.a` for the `$(TARGET)` without `DDISABLE_CALL_ONCE_STATIC`, and one `librdma_for_plugin.a` for `$(PLUGIN_SO)` with the macro. The reason is: I think `DISABLE_CALL_ONCE_STATIC` will severely impact the load balancing among engines because of a non-static `std::vector<std::atomic<uint32_t>>* off` that causes all connections will map to the first engine.

p2p/engine.cc
```diff
@@ -34,7 +34,9 @@ inline void check_python_signals() {
 
 Endpoint::Endpoint(uint32_t const local_gpu_idx, uint32_t const num_cpus)
     : local_gpu_idx_(local_gpu_idx), num_cpus_(num_cpus) {
-  py::gil_scoped_release release;
+  if (PyGILState_Check()) {
+    py::gil_scoped_release release;
```
YangZhou1997: This will immediately get destroyed inside the `{}`, similar for others. Suggesting using `[[maybe_unused]] auto _ = xxx`

p2p/Makefile
praveingk: Makes sense 👍🏽 Will do that.

p2p/engine.cc
praveingk: Done

p2p/Makefile
praveingk: Done

Issue #194, opened 2025-07-24
Using pip to automatically manage all external .so files


YangZhou1997:
Ideally, any lib should be installed by `pip install -e .`

This is especially important for some AMD testbeds that do not provide root access to install packages freely, eg, AMD version of `rdma/` and `p2p`

YangZhou1997:
Using docker build in https://github.com/uccl-project/uccl/pull/207

PR #195, opened 2025-07-24
make p2p work on amd

YangZhou1997:
I also commented out `#define USE_REDIS` in engine.h to make the build easier on AMD. @MaoZiming FYI. Later, I will figure out a way to install all deps via pip. See https://github.com/uccl-project/uccl/issues/194. 

image


MaoZiming:
Sounds good (commenting out redis)

PR #202, opened 2025-07-25
GPU p2p on gh200 now works using ibv_reg_mr_iova2

YangZhou1997:
Unrelated to these changes, I find sometimes the p2p server side will trigger errors: 

image

@zhongjiechen @MaoZiming any thoughts?

### DIFF
--- a/p2p/README.md
+++ b/p2p/README.md
@@ -102,14 +102,13 @@ python benchmark_uccl.py --role server --device gpu --local-gpu-idx 0 --num-cpus
 
 On client:
 ```bash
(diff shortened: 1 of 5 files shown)


YangZhou1997:
Return code is 12: IBV_WC_RETRY_EXC_ERR

MaoZiming:
Oh observed this error too.

zhongjiechen:
FYI: `ureq->recv.qp` is the per-connection RC QP for coordinating buffers between receiver and sender.

MaoZiming:
UCCL read benchmark works. I wonder if this bug is because of the chunking

YangZhou1997:
fixed by making more frequent poll_flow_cq().

PR #207, opened 2025-07-26
refine docker build for p2p

YangZhou1997:
Deprecating pip install, as it is still too early for us, eg, sometimes EFA lib broken, and we are releasing new libs like P2P. 

Because we have many dependencies on the platform and thirdparty libs, I have not figured out a way to just `pip install -e .`, let's use docker build and install for now. 

```
git clone https://github.com/uccl-project/uccl.git --recursive
bash docker_build_and_install.sh [cuda|rocm] [py_version]
```


YangZhou1997:
Note that now, we do not fully support cross-platform wheel building. For example, we cannot build rocm on ARM platform, and cannot build efa .so on AMD platform. 

We suggest building publishable x86 wheels on either cuda+x86 or rocm+x86 platform, arm64 wheels on GH200 platform.

YangZhou1997:
Once uploaded to pypi, you can install cuda version by `pip install uccl`, or rocm version by `pip install uccl==1.0.0+rocm`

PR #208, opened 2025-07-26
Add Nvlink copy for p2p

MaoZiming:
Adding a simple API for local inter-process `cudaMemcpy`.
UCCL P2P will initiate a dummy connection `kNvlinkConn`.
Current code uses out-of-band notification for completed transfer. In a future PR, this should be folded into UCCL and UCCL should provide a receiver API for application to poll for completion, similar to `Endpoint::recv`.

### DIFF
--- a/include/util/cuda.h
+++ b/include/util/cuda.h
@@ -1,0 +1,12 @@
+#pragma once
+#include <cuda_runtime.h>
+
+#define CUDA_CHECK(call)                                            \
+  do {                                                              \
+    cudaError_t err__ = (call);                                     \
+    if (err__ != cudaSuccess) {                                     \
+      fprintf(stderr, "CUDA error %s:%d: %s\n", __FILE__, __LINE__, \
+              cudaGetErrorString(err__));                           \
+      std::abort();                                                 \
+    }                                                               \
+  } while (0)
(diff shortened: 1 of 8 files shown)


MaoZiming:
@YangZhou1997  APIs need some discussion.
Currently:
| Sender                             |                             Receiver                             |
|------------------------------------|------------------------------------------------------------------|
|      `endpoint()`            | `endpoint()`                                                     |
|                                    | oob send endpoint metadata                                   |
| oob receive endpoint metadata  |                                                                  |
| `endpoint.connect(metadata)`               |                                                                  |
| → `conn_id = kNvlinkConn` (detected same server, different GPU via NVLink) |                          |
|                                    | `advertise(kNvlinkConn, tensor)` → returns blob (**cudaIpcMemHandle**)                 |
|                                    | oob send blob                                                |
| oob receive blob              |                                                                  |
| `endpoint.send()`  → calls `cudaMemcpy` because  `kNvlinkConn`              |                                                                  |
|                                    | no need for `recv()` on receiver |

YangZhou1997:
Sounds great. Just a quick question: since there is no need for recv() on the receiver, how could the receiver know the transfer has been finished? There might still be some poll() function on the receiver side, I guess, even for pull-based APIs (so that the pull target can safely release the tensor). 

Another note is: let's try to use more handy `torchrun` with `pytorch.dist("gloo")` for OOB, see `benchmark_uccl.py`

MaoZiming:
@YangZhou1997 I wonder if UCCL should manage a small flag for receiver to poll, or the application should do that instead.

PR #212, opened 2025-07-28
simplify recv logic by assuming recv size is know OOB

YangZhou1997:


YangZhou1997:
On AMD GPUs + Thor-2 NICs: 

UCCL P2P previous: 
image

UCCL P2P now:
image

RCCL: 
image

YangZhou1997:
@MaoZiming , I change `include/util/cuda.h` to `include/util/gpu_rt.h` to abstract away CUDA and HIP details with some macros. 

One weird thing is that on AMD GPUs, any hip calls (such as hipStreamCreateWithFlags) will give the error of "no ROCm-capable device is detected"

YangZhou1997:
Oh, I figure it out the reason of "no ROCm GPUs": I need to further exclude "libamdhip64.so.6" when building wheels to let the p2p .so link to the sys-install libamdhip64.so.6

YangZhou1997:
On GH200 + CX-7 single port: 

UCCL p2p: 
image

NIXL:
image

NCCL: 
image

PR #215, opened 2025-07-28
Add gpu_driven_amd with AMD GPU support

CalebZ9909:
Uploaded first version of converted gpu_driven that can work on AMD GPUs


MaoZiming:
Hi @CalebZ9909 , great work! could you also bring back the deleted files? I am seeing 530 changed files. It would be easier to review the code that way. Thank you!

PR #217, opened 2025-07-28
Add gpu_driven_amd with AMD GPU support

CalebZ9909:
Add AMD HIP for gpu_driven on AMD GPUs

### DIFF
--- a/gpu_driven_amd/Makefile
+++ b/gpu_driven_amd/Makefile
@@ -1,0 +1,90 @@
+ABS_ROOT_PATH := $(shell dirname $(realpath $(lastword $(MAKEFILE_LIST))))
+
+RCCL_HOME:=../thirdparty/rccl
+HIP_HOME?=/opt/rocm
+CONDA_LIB_HOME?=/usr/lib
+
+INC = -I./ -I../include -I$(HIP_HOME)/include -I${CONDA_LIB_HOME}/../include -L${CONDA_LIB_HOME}
+LIBS = -lz -lelf -libverbs -L ${HIP_HOME}/lib -lamdhip64
+LIBS_SHARED = -lglog -lgflags -lgtest -lz -lelf -libverbs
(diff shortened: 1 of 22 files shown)


YangZhou1997:
@CalebZ9909 Looks great. Can you put your code also in the `gpu_driven` folder so that we can track the diff and do the merging?

YangZhou1997:
Nice. I think my past wording was confusing. You should also do `git rm -r gpu_driven_amd`.

CalebZ9909:
Yes, I am done on my side. Ziming and you can check the code now. I am working on testing the code and will update next version soon.

MaoZiming:
Thanks @CalebZ9909 , ideally we can define a flag (e.g. __HIP_PLATFORM_AMD__ in rdma folder) so that we can toggle whether to run on AMD platform from the makefile. It should not overwrite all existing cuda commands with hip but keeping the option to toggle between the two

Maybe we can define some function name to either refer to the cuda version or the hip version depending on the macro. e.g. (either hipMemcpyPeerAsync or cudaMemcpyPeerAsync).

YangZhou1997:
I have some macro defination at https://github.com/uccl-project/uccl/blob/main/include/util/gpu_rt.h

CalebZ9909:
Sounds good! Yes, I can set up a flag and keep all CUDA-related files along with AMD files, which can be chosen to be executed by the flag in the makefile.

YangZhou1997:
I guess a better way is to use the unified name like `gpuMemcpy` in https://github.com/uccl-project/uccl/blob/main/include/util/gpu_rt.h, so that we just need to write a new `MakefileHip` for AMD. You can find an example in https://github.com/uccl-project/uccl/tree/main/p2p

CalebZ9909:
Sure, I will look into this!

MaoZiming:
This is already merged I believe. closing the PR.

PR #218, opened 2025-07-28
benchmarking read, migrating read to use torchrun

YangZhou1997:


YangZhou1997:
I also figure out why https://github.com/uccl-project/uccl/pull/180#issuecomment-3104934750 finds `benchmark_nixl2.py` (one shared NIXL agent with multiple transfer_handles) slow. It only happens when using cpu memory, but not happening when using gpu memory on GH200.

PR #220, opened 2025-07-28
Add dual proxy to gpu-driven

MaoZiming:
After https://github.com/uccl-project/uccl/pull/211

### DIFF
--- a/gpu_driven/bench/benchmark_dual.cu
+++ b/gpu_driven/bench/benchmark_dual.cu
@@ -13,7 +13,8 @@
 
 int main(int argc, char** argv) {
   if (argc < 3) {
-    std::cerr << "Usage: ./benchmark_dual <rank> <peer_ip>\n";
+    std::cerr << "Usage: CUDA_MODULE_LOADING=EAGER ./benchmark_dual <rank> "
+                 "<peer_ip>\n";
     return 1;
   }
   int const rank = std::atoi(argv[1]);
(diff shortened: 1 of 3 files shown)


MaoZiming:
The code is a bit flaky.

MaoZiming:
@YangZhou1997 
This current works.
One interesting is that under dual mode, launching a kernel to do thread copy seems to contend with user application kernel, causing it to get stuck in some scenarios. I am not entirely sure why. This needs to be considered again in the future.
I switched back to `cudaMemcpyPeerAsync`. It seems to be fine.

YangZhou1997:
Kernel launches do interfere with each other, even if they are from two streams. This is because of the lazy loading of cuda functions which requires a global device sync, see https://forums.developer.nvidia.com/t/lazy-loading/261135 for details. This gives troubles especially when handling persistent kernels. 

You turn lazy loading off by `CUDA_MODULE_LOADING= EAGER`. See some of my experiments here: https://github.com/uccl-project/uccl/blob/main/misc/cuda_persist_kernel.cu

YangZhou1997:
Maybe it is time to make it work on AMD GPUs? with yihan and chihan. I find the 400Gbps AMD cluster pretty useful once we make our code work on it by writing a MakefileHip.

MaoZiming:
@YangZhou1997 For some reason, it crashes with `CUDA_MODULE_LOADING=EAGER`

MaoZiming:
@YangZhou1997 Thanks. that seems to be the issue!

PR #223, opened 2025-07-29
Add an initial copy-only ccl for sendrecv with pytorch.dist like APIs

YangZhou1997:


YangZhou1997:
`benchmark_uccl.py`: 
image

`benchmark_uccl_collective.py`: 
image

`benchmark_nccl.py`: 
image

PR #236, opened 2025-07-31
Initial attempt to translate gpu_driven to amd. Two parts left: __builti…

YangZhou1997:
…n_nontemporal_load and cuda_pipeline


MaoZiming:
Thanks! will take a look tomorrow

YangZhou1997:
Yihan has a pipelined copy kernel for AMD: https://github.com/CalebZ9909/uccl/blob/ec91df9a9bfd4a824bf05b5cb0fd92127e82c488/gpu_driven/src/peer_copy.hip#L123, we should merge it

`__builtin_nontemporal_load` usage is basically borrowed from rccl `thirdparty/rccl/src/device/op128.h`, but it does not get compiled in MakefileHip

YangZhou1997:
Now it compiles successfully!

HermesCui:
I fixed 3 issues:

RingBuffer initialization issue: Uninitialized head/tail contained garbage values (e.g., head=18446744072627423018), causing popN() to calculate incorrect available data and access invalid memory; fixed by explicitly initializing to 0 to ensure a clean starting state.

Type conversion issue: Assigning popN()'s int return value to size_t caused negative values to be converted to huge positive numbers (e.g., 18446744071582163775) used as array indices, triggering segfaults; fixed by using int type to avoid unsigned conversion problems.

Iteration issue: The issue was that the sender was transmitting data much faster than the receiver could process it, causing the GPU kernel to stall while waiting for the receiver to catch up. With kIterations set to 40,000, the benchmark runs to completion without any bottlenecks.

YangZhou1997:
@HermesCui, @CalebZ9909 , this is great! @MaoZiming , can any of you try to also test if the code using Makefile can work on the GH200? If so, we should merge this PR quickly. 

Another PR might be that we need some back-pressure mechanism to control the load to the GPU kernel. 
One quick suggestion is to have a dedicated folder called `kernels` or `device` to hold all the kernel code. Ideally, the kernel code should be decoupled from the CPU code to avoid compilation issues and evolve them separately.

CalebZ9909:
Yes, it works on the GH200 node now! I just fixed some compile issues. I will push soon, note that uccl/gpu_driven/include/common.hpp, we need to set up the #define NUM_GPUS based on the testbed. Maybe mention this in the README?

MaoZiming:
@CalebZ9909 Great! Feel free to add that to the readme

MaoZiming:
I think this should be `cudaFreeHost`. pushed a commit to fix. 
```
include/bench_utils.hpp(50): error: identifier "cudaHostFree" is undefined
      do { cudaError_t err__ = (cudaHostFree(env.rbs)); if (err__ != cudaSuccess) { fprintf(
                                ^

include/bench_utils.hpp(96): error: identifier "cudaHostFree" is undefined
    do { cudaError_t err__ = (cudaHostFree(p)); if (err__ != cudaSuccess) { fprintf(
                              ^
```

gpu_driven/bench/benchmark_dual.cu
```diff
@@ -3,6 +3,7 @@
 #include "peer_copy_worker.hpp"
```
MaoZiming, line 3: Can we merge the `cu` and `hip` benchmark files

gpu_driven/src/peer_copy.cu
```diff
@@ -109,9 +109,10 @@ cudaError_t launch_peer_bulk_copy2(CopyTask const* host_tasks, int num_tasks,
         <<<blocks, threads_per_block, shmem, stream>>>(d_tasks, num_tasks,
                                                        tasks_per_block);
   }
-  return cudaGetLastError();
+  return gpuGetLastError();
 }
 
+#ifndef __HIP_PLATFORM_AMD__
 template <int PIPE_DEPTH = 2, typename VecT = int4>
 __global__ void peer_copy_kernel_vec_pipelined(
```
MaoZiming, line 119: This is not the fastest on nvidia so it won't be executed.
https://github.com/uccl-project/uccl/blob/main/gpu_driven/src/peer_copy.cu#L105-L111

gpu_driven/bench/benchmark_dual.cu
YangZhou1997: I am using software links for .hip benchmarks, as it seems hipcc requires the source file to be in .hip extension. Maybe there are some flags that one can set for hipcc to use .cu?

YangZhou1997:
@MaoZiming @HermesCui @CalebZ9909 We should be able to merge right? since it works on both GH200 and AMD testbeds. We can open future PRs to further improve it.

gpu_driven/src/peer_copy.cu
YangZhou1997: Let's keep it for now, incase future may need it.

MaoZiming (approved):
LGTM

CalebZ9909:
Lgtm too

PR #237, opened 2025-08-01
add dpdk driver for uccl

Nelson-Cheung:


YangZhou1997:
Thank you for your effort @Nelson-Cheung! Is this already running on cloudlab? maybe write a readme for how to run it?

Nelson-Cheung:
I am writing `util_dpdk_test.cc` and modifying `transport.cc`. If both are run through, there will be no problem.

> Thank you for your effort @Nelson-Cheung! Is this already running on cloudlab? maybe write a readme for how to run it?

PR #244, opened 2025-08-03
transparently handle rcmode

YangZhou1997:


YangZhou1997:
Now the logic for choosing UC vs RC is: 

if RCMODE set: 
   use RC
   if support_uc: print out a warning on "less scalable perf"
else:
   if support_uc: use UC
   else: use RC

@MaoZiming , do we still need to do this check now? 
image

MaoZiming:
@YangZhou1997 How is current read API implemented if not on RC? iiuc UCCL read with RDMA read does not work if not RC?

YangZhou1997:
Yes, RDMA read requires RC. But it seems that the current `uccl_read_async` always uses RC QP (this RC QP was for small message bypassing UCCL multipath): 

image

MaoZiming:
The current code only triggers `rc_read` when `RDMAEndpoint::RC_MAGIC` is set. Should we call `rc_read` from endpoint directly?

YangZhou1997:
I just tested and find even if not specifying rcmode, the benchmark_uccl_read.py still works

YangZhou1997:
Oh, I think I got what you mean; I will still let the read API require UCCL_RCMODE. Ideally, we want to be able to use UC send and RC read at the same time.

YangZhou1997:
Btw, I find read perf is indeed bad at 400G AMD servers: 

`benchmark_uccl_read.py`
image

PR #246, opened 2025-08-04
Optimizing uccl collective performance

YangZhou1997:

PR #247, opened 2025-08-04
Make UCCL run Primus.

zhongjiechen:


zhongjiechen:
We should **never use `get_best_dev_idx()` in nccl_plugin.cc**, but let NCCL get the PCIe path through `pluginGetProperties()`.

This is because the GPU RANK ID obtained from `hipGetDevice()` in `nccl_plugin.cc` is different the GPU order from:
https://github.com/uccl-project/uccl/blob/main/rdma/rdma_io.cc#L67-L72

UCCL's current automatic detection works well.

YangZhou1997:
Hit the same problem for p2p engine (https://github.com/uccl-project/uccl/pull/246), I will think of a better way of mapping GPU rank to GPU idx and RDMA NIC idx.

YangZhou1997:
It should has been fixed in https://github.com/uccl-project/uccl/pull/246, get_best_dev_idx() should now accept input rank got from cudaGetDevice()

zhongjiechen:
In some cases (Primus w/ nvlink on), `pluginListen()` is not called before calling `pluginConnect()` on a dev, so we need to call `initialize_engine_by_dev()` in `pluginConnect()`.

https://github.com/uccl-project/uccl/blob/Primus/rdma/nccl_plugin.cc#L351-L357

zhongjiechen:
Support for NIC sharing. The modification is very simple, just set the peer id as a variable in each process and set the remote GPU index as a key in peer_map.

https://github.com/uccl-project/uccl/blob/Primus/rdma/transport.cc#L1054-L1056

YangZhou1997:
GPU seems all full, I guess it creates too many UCCL engines. 

image

YangZhou1997:
I feel UCCL somehow launches too many engine threads, see this screenshot of CPU usage. Probably `init_dev[]` might cross different processes? 

image

zhongjiechen:
I think so too. I found in the logs that each GPU was using two NICs, and there was no corresponding `listen()` before the `connect()` call, so I supported them in this PR. But now I think this may be caused by a topology error?

YangZhou1997:
Debugging notes: 
* when NVLink is off, everything is good, but when NVLink is on, each of the eight processes will call `ep->initialize_engine_by_dev` for each of the eight RDMA devices. I find that these 8x8 initiations do not happen to RCCL. 
* by forcily overwritting `dev=local_gpuidx`, UCCL still works but just less efficient than RCCL -> is this caused by small message size so that UCCL inherently performs worse? need to print out the message size. 
* RCCL has some early termination for listen and connect---do we also need that? 
  * image

zhongjiechen:
> by forcily overwritting `dev=local_gpuidx`, UCCL still works but just less efficient than RCCL -> is this caused by small message size so that UCCL inherently performs worse? need to print out the message size.

Just to confirm, does this produce the correct mapping with the latest code? In the previous code, `local_gpuidx = 0` gets the GPU that is not the closest to rdma0.

YangZhou1997:
Oh, it does not. That reminds of what if we align the order of rdma_ctrl->devices_ with GPU order (most affinitied). I just implemented it and find this brings performance back: on two nodes. 

UCCL: 140.3/141.2 ms vs. RCCL 135.7/138.5 ms

This is done when `dev=ep->get_best_dev_idx(local_gpuidx)` (I also fixed a gpu-dev mapping bug).

YangZhou1997:
Oh, I see the reason. I did more thorough profiling on RCCL `net_ib.cc`, and conclude that RCCL `net_ib.cc` will indeed, for each GPU process, call `ncclIbConnect` for each RDMA dev (see below). This is likely for PXN to let each process manage other NICs (even if we disable PXN, this still happen probably because of forgetting checking on PXN during initing phase).

image 

So we have two solutions: 
* implementing cross-process support
* overwritting `dev=ep->get_best_dev_idx(local_gpuidx)` and ignoring PXN at the plugin level. 

I currently go for solution 2, which works for both Primus and nccl-tests (1process1GPU and 1process8GPUs)

YangZhou1997:
Hi @zhongjiechen , do you think we can merge this PR, while leaving Primus parameter optimization for later PR (global bz and micro bz)?

zhongjiechen:
LGTM!

PR #252, opened 2025-08-05
Running nixl on amd+broadcom

YangZhou1997:


YangZhou1997:
NIXL results (single direction): 
image

UCCL P2P results (single direction): 
image

NIXL read: 
image

MaoZiming:
@YangZhou1997 It seems NIXL is quite good on AMD

YangZhou1997:
@MaoZiming Yes, it is pretty good. I was thinking of two directions: 
* Multi-device (8GPUs to another 8GPUs, NIXL might just use one thread.)
* Dual-directional (again, NIXL might just use one thread)
* Copy-only CCL (so our baseline is NCCL)

All in all, RDMA P2P transfer without any network congestion is already pretty optimized.

PR #255, opened 2025-08-06
Add UCCL proxy initialization before dispatch/combine buffer

MaoZiming:


gpu_driven/bench/test_internode_simple.py
```diff
@@ -44,12 +49,29 @@ def test_simple_internode(rank: int, num_ranks: int, group: dist.ProcessGroup):
...
+            rb_addr=bench.ring_addr(i),
+            block_idx=i,
+            gpu_buffer_addr=buf_ptr,
+            total_size=bytes_needed,
+            rank=rank,
+            peer_ip=peer_ip,
+        )
+        proxy.start_dual()
+        proxies.append(proxy)
+    ep.register_proxy(device_index, proxies[0])
```
MaoZiming: This is outside the Buffer API, but I think it's needed for UCCL.

MaoZiming: @YangZhou1997

PR #257, opened 2025-08-06
quick fix on shared_pool

YangZhou1997:


YangZhou1997:
During `~ThreadCache()` in shared_pool.h, it needs to return back its local cached items to the global pool. However, previously, I forgot to add the global lock.

PR #259, opened 2025-08-06
Add `buffer.py` and bring torch to uccl_ep.cc

MaoZiming:
* `uccl_ep.cc` mirrors `deep_ep.cc`
* Adding `buffer.py`, similar to `buffer.py` in `thirdparty/DeepEP/deep_ep/buffer.py`.
* Added Torch dependencies in Docker file.


MaoZiming:
Use `./build_and_install.sh cuda all 3.11`

PR #262, opened 2025-08-08
removing efa, misc, rdma usage of dedicate hostfile

YangZhou1997:


YangZhou1997:
We can use `--map-by ppr:${PROCS_PER_NODE}:node` in mpirun to choose how many processes per node.

PR #269, opened 2025-08-10
add dual directional test for nixl

YangZhou1997:


YangZhou1997:
NIXL dual direction: 

image

As a reference, UCCL P2P dual direction: 

image

PR #274, opened 2025-08-10
P2p collective

YangZhou1997:


YangZhou1997:
There are some (likely) data consistency issues. Need to check how NCCL handles it. 

`torchrun --nnodes=1 --nproc_per_node=2 --node-rank=0 benchmark_uccl_collective.py --ring`

You can add `--log-dir=./logs --redirect=3` to redirect torchrun logs.

YangZhou1997:
Oh wait, it still has some value errors.

YangZhou1997:
Okay, interesting. It seems that `hipIpcGetMemHandle()` will auto-transform 512B-aligned address into some higher-aligned address. By default, pytorch tensor is allocated with 512B alignment. 

For example, a source data address of `0x7f781c400200` on the receiver side becomes `0x7f0f9bc00000` when opened the `IpcMemHande` at the sender side. 

So the final fix is also easy, just forcely pass offset and reconstruct the 512B-aligned address at the other side.

YangZhou1997:
Our P2P performance via DMA engine on AMD: 
image


As a reference, RCCL 2-GPU sendrecv: 
image

Issue #276, opened 2025-08-12
Question about NIC and GPU support


cos120:
Hi, thanks for great project!

I try UCCL with 2 nodes, each of node has 8*A100 and 4 of dual-port ConnectX-6 Dx, I build with commit id a2b6bbd3f51dbea8f5ef0376d580d62b772ed6d2. When I run all_gather via pytorch, it abort.

```bash
F0812 05:53:36.518041 3243288 rdma_io.h:570] Check failed: it != rdma_ctl->gpu_to_dev_idx_.end() Error: gpu_to_dev_idx_[4] is not set.
F0812 05:53:36.517851 3243281 rdma_io.h:570] Check failed: it != rdma_ctl->gpu_to_dev_idx_.end() Error: gpu_to_dev_idx_[6] is not set.
```

rdma_io.h:570 is 
```cpp
static inline int get_best_dev_idx(int gpu_idx) {
  auto it = rdma_ctl->gpu_to_dev_idx_.find(gpu_idx);
  CHECK(it != rdma_ctl->gpu_to_dev_idx_.end())
      << "Error: gpu_to_dev_idx_[" << gpu_idx << "] is not set.";
  return it->second;
}
```

My pcie topo

```bash
 |           \-02.0-[a6-c7]----00.0-[a7-c7]--+-00.0-[a8-ae]----00.0-[a9-ae]--+-00.0-[aa]--
 |                                           |                               \-10.0-[ab-ae]----00.0-[ac-ae]--+-00.0-[ad]----00.0  NVIDIA Corporation GA100 [A100 SXM4 80GB]
 |                                           |                                                               \-1f.0-[ae]----00.0  Broadcom / LSI PCIe Switch management endpoint
 |                                           +-04.0-[af-b4]----00.0-[b0-b4]--+-00.0-[b1-b3]----00.0-[b2-b3]----00.0-[b3]----00.0  NVIDIA Corporation GA100 [A100 SXM4 80GB]
 |                                           |                               \-10.0-[b4]--+-00.0  Mellanox Technologies MT2892 Family [ConnectX-6 Dx]
 |                                           |                                            \-00.1  Mellanox Technologies MT2892 Family [ConnectX-6 Dx]
```

Looks like UCCL can not find nic index for rank 4 and rank6? Does UCCL can support that 2 gpu sharing one nic or some special config? Thanks!

praveingk:
[...]

PR #277, opened 2025-08-12
Add normalization for GPU ID

YangZhou1997:


YangZhou1997:
Related to https://github.com/uccl-project/uccl/issues/276, that `cudaDeviceGetPCIBusId` may return upper case characters.

Issue #267, opened 2025-08-09
ibv_post_send return error 12


YangZhou1997:
On the new AMD cluster, there seems to be no network congestion at all. So we set `UCCL_ENTROPY=1`, but this will trigger ibv_post_send error in supply_rx_buff() for 1GB messgaes: https://github.com/uccl-project/uccl/blob/96d71231f543056e19f6c87e6cc9267eab8efebc/rdma/transport.cc#L1987-L1988

You can reproduce the bug by `https://github.com/uccl-project/uccl/blob/main/rdma/run_chi.sh`

@zhongjiechen, any thoughts on this? (this is low priority, as in the new AMD cluster with no congestion, we have no gains on collectives)

zhongjiechen:
I have encountered three situations that will cause `ibv_post_send()` to return 12 so far: 
- (1) inline is specified, but the data size exceeds the inline size
- (2) the number of sge is greater than the set max_send_sge
- (3) cqes are not polled timely, resulting in SQ overflow

(3) is more likely, the value here seems too small (16 * 1):
https://github.com/uccl-project/uccl/blob/96d71231f543056e19f6c87e6cc9267eab8efebc/rdma/transport.h#L1361

YangZhou1997:
It seems that after tuning `kMaxRecv * kMaxReq` or kMaxSendRecvWR to 4x of the previous value (https://github.com/uccl-project/uccl/pull/283/files). The `IBV_WC_RETRY_EXC_ERR` error does not happen anymore on the new AMD cluster (20 runs).

PR #283, opened 2025-08-15
tune some parameters for congestion-free clusters

YangZhou1997:


YangZhou1997:
Trying to fix https://github.com/uccl-project/uccl/issues/267

YangZhou1997:
Most RCCL runs are pretty good: 
image

RCCL's results: 
image

To replicate: 
* `./run_chi.sh uccl`
* `./run_chi.sh rccl`

Note that the third node does not seem to be connecting to others; even RCCL cannot run with it. 

Sometimes (not too common) for large messages, UCCL will get low perf. Any thoughts @zhongjiechen 
image

YangZhou1997:
Well, it seems RCCL will also hit similar issues sometimes (32MB). @zhongjiechen , I guess we can ignore this suddle drop for UCCL now: 

image

PR #286, opened 2025-08-16
Fix the remote gpu addr translation without nvshmem

CalebZ9909:


CalebZ9909:
@MaoZiming Hi Ziming, you can review here, we can keep this pr for tracking deploying the cpu_proxy internode_low_latency kernel on my side

CalebZ9909:
@MaoZiming @YangZhou1997 Hi All, Based on the results of running simple internode test, the UCCL_DeepEP dispatch and combine work now. I checked the remote address and offset, and they look good to me. One flaw is that the test does not quit normally; it seems that all ranks are not well synced at the end. I will fix this soon.

MaoZiming:
@CalebZ9909 Thanks! Great work! I just got back, will review in the afternoon!

gpu_driven/include/uccl_ibgda.cuh
```diff
@@ -41,12 +41,12 @@ __device__ __forceinline__ void nvshmemi_ibgda_put_nbi_warp(
...
-      "dst_rank: %d, sm_id: %d, lane_id: %d, message_idx: %d, num_ring_addrs: "
-      "%d, cur_head: %llu, cur_tail: %llu, inflight: %llu\n",
-      rptr_val, lptr_val, bytes_val, dst_rank, sm_id, lane_id, message_idx,
-      num_ring_addrs, cur_head, cur_tail, inflight);
+  // printf(
+  //     "[ibgda_put_nbi_warp] req_rptr: %llu, req_lptr: %llu, bytes: %llu, "
+  //     "dst_rank: %d, sm_id: %d, lane_id: %d, message_idx: %d, num_ring_addrs: "
+  //     "%d, cur_head: %llu, cur_tail: %llu, inflight: %llu\n",
+  //     rptr_val, lptr_val, bytes_val, dst_rank, sm_id, lane_id, message_idx,
+  //     num_ring_addrs, cur_head, cur_tail, inflight);
```
MaoZiming, line 49: We will remove these printf after the code works.

gpu_driven/src/rdma.cpp
```diff
@@ -456,34 +458,27 @@ void post_rdma_async_batched(ProxyCtx& S, void* buf, size_t bytes,
...
+    // TODO(Ziming): review the logic of following comment here
+    // ORIGINAL CODE: Incorrect address calculation ignoring cmd.req_rptr
     // wrs[i].wr.rdma.remote_addr = cmd.req_rptr ? cmd.req_rptr : S.remote_addr;
-    // TODO(MaoZiming): fix address translation.
-    // NOTE(MaoZiming): cmd.req_rptr is not correct. Because uccl_ep has not yet
-    // done rdma handshake, so it doesn't know the remote address.
-    wrs[i].wr.rdma.remote_addr = S.remote_addr + i * bytes;
+    // wrs[i].wr.rdma.remote_addr = S.remote_addr + i * bytes;
+    
+    wrs[i].wr.rdma.remote_addr = S.remote_addr + cmd.req_rptr;
```
MaoZiming, line 466: @CalebZ9909 Can you double check whether this offset (`cmd.req_rptr`) is with respect to `dispatch_rdma_recv_data_buffer` in `LowLatencyBuffer`? 
This might not be the correct offset since `S.remote_addr` stores the start of the `LowLatencyBuffer buffers[2]`

MaoZiming: Similarly for `nvshmemi_ibgda_amo_nonfetch_add`, need to check the offset of `dispatch_rdma_recv_count_buffer` with respect to `rdma_buffer` in `LowLatencyLayout`.

MaoZiming: For reference
https://github.com/uccl-project/uccl/blob/main/gpu_driven/include/ep_config.hpp#L107-L125

CalebZ9909: Sure, will modify this

PR #287, opened 2025-08-19
Eliminate Nvshmem get p2p

HermesCui:
Eliminated uccl::nvshmemi_get_p2p_ptr(dst_ptr, rank, dst_rank);


gpu_driven/src/internode_ll.cu
```diff
@@ -184,13 +184,18 @@ __global__ __launch_bounds__(1024, 1) void dispatch(
...
         }
         
-#ifdef false
-        auto const dst_p2p_ptr =
-            uccl::nvshmemi_get_p2p_ptr(dst_ptr, rank, dst_rank);
-        if (dst_p2p_ptr == 0) {
-          uccl::nvshmemi_ibgda_put_nbi_warp(dst_ptr, src_ptr, num_bytes_per_msg,
-                                            dst_rank, dst_expert_local_idx,
-                                            lane_id, slot_idx);
+#ifdef true
```
MaoZiming: Feel free to delete these `#ifdef true` if they work

gpu_driven/src/internode_ll.cu
```diff
@@ -292,12 +297,15 @@ __global__ __launch_bounds__(1024, 1) void dispatch(
...
+    // auto dst_p2p_ptr = nvshmemi_get_p2p_ptr(dst_ptr, rank, dst_rank);
+    if (dst_rank != rank) {
+      // uccl::nvshmemi_ibgda_amo_nonfetch_add(reinterpret_cast<int*>(dst_ptr),
+      //                                       -num_tokens_sent - 1, dst_rank,
+      //                                       dst_expert_local_idx);
+      uccl::nvshmemi_ibgda_amo_nonfetch_add(
+        reinterpret_cast<int*>(dst_offset), -num_tokens_sent - 1, dst_rank, sm_id,
+        dst_expert_local_idx, false, ring_addrs, num_ring_addrs);
     } else {
       st_release_sys_global(reinterpret_cast<int*>(dst_p2p_ptr),
```
MaoZiming, line 284: Does `st_release_sys_global` still work for `dst_p2p_ptr`?

HermesCui: It should not work it, but our current test does not trigger these {else}s.

HermesCui: I will fix the issue later, we need to directly calculate the dst_ptr

MaoZiming: Sounds good. We should have test that trigger intra-node transfers.

gpu_driven/src/internode_ll.cu
```diff
@@ -792,11 +800,14 @@ __global__ __launch_bounds__(1024, 1) void combine(
         ;
       auto dst_ptr =
           reinterpret_cast<uint64_t>(rdma_recv_flag + global_expert_idx);
-#ifdef false
-      auto dst_p2p_ptr = nvshmemi_get_p2p_ptr(dst_ptr, rank, dst_rank);
-      if (dst_p2p_ptr == 0) {
+#ifdef true
+      // auto dst_p2p_ptr = nvshmemi_get_p2p_ptr(dst_ptr, rank, dst_rank);
+      if (dst_rank != rank) {
```
MaoZiming: if (`dst_rank != rank`), we might eventually need a way to tell (maybe from rank?) which ranks are on the same node. For these ranks, go through the intra-node path.

HermesCui: auto const dst_rank = dst_expert_idx / num_local_experts;
auto const dst_expert_local_idx = dst_expert_idx % num_local_experts;
According to the above code, dst_rank and rank represent node IDs (Not specific GPU). so we can use them to determine if an operation is inter or intra I feel that we don't need to know which specific GPU each rank belongs to.

MaoZiming: Sounds good

MaoZiming: @HermesCui 
Re: dst_rank and rank represent node IDs (Not specific GPU). 
It seems each GPU gets a different rank? 
`num_local_experts` is the number of experts per GPU? 
See: `num_local_experts = num_experts // num_ranks`

MaoZiming: e.g. in the intra-node test, https://github.com/deepseek-ai/DeepEP/blob/main/tests/test_intranode.py#L15C72-L15C81
you can have `num_ranks = 8`

HermesCui: Oh, I remember that their test_intranode test calls intranode.cu, not internode_ll.cu. For intranode.cu, they assume that DeepEP will run on a single node with 8 GPUs.

HermesCui:
1. Replaced NVSHMEM dependency: Implemented `get_ipc_p2p_ptr() ` to determine intra-node vs inter-node communication without NVSHMEM
2. IPC handle exchange: Host creates and all-gathers IPC handles for GPU buffers, opening handles for intra-node GPU pairs
3. API extension: Added IPC pointer array parameter to support direct intra-node GPU memory access

gpu_driven/include/ep_configs.cuh
```diff
@@ -1,6 +1,6 @@
 #pragma once
 
-#define NUM_MAX_NVL_PEERS 8
+#define NUM_MAX_NVL_PEERS 1
```
MaoZiming: Can we leave this as 8 _(maximum),_ and when we need to check the number of devices to talk to, get the number of devices dynamically with `cudaGetDeviceCount`?

gpu_driven/src/proxy.cpp
```diff
@@ -98,6 +98,7 @@ void Proxy::run_dual() {
 
   uint64_t my_tail = 0;
   size_t seen = 0;
+  auto last_print = std::chrono::steady_clock::now();
```
MaoZiming: We don't need this?

gpu_driven/include/ep_configs.cuh
HermesCui: yeah, I just set it to 1 for testing one gpu node.

gpu_driven/src/proxy.cpp
HermesCui: We can keep it for future testing.

MaoZiming:
LGTM. By the way, we should minimize the configuration needed for parameters, especially if they can be easily detected (https://github.com/uccl-project/uccl/issues/311) but this can be left to the future.

gpu_driven/include/ep_configs.cuh
MaoZiming: I have added a new commit to set this dynamically.

PR #289, opened 2025-08-20
uccl_deepep address passing & atomic add

CalebZ9909:


CalebZ9909:
@MaoZiming Hi Ziming, the CPU GPU offset mismatch is solved now by setting a dispatch_rdma_recv_data_buffer in rdma, which can be set in proxy

gpu_driven/bench/test_internode_simple.py
```diff
@@ -104,6 +104,31 @@ def test_simple_internode(rank: int, num_ranks: int, group: dist.ProcessGroup):
...
+        
+        dispatch_recv_count_buffer_bytes = num_experts * 4  # sizeof(int)
+        signaling_buffer_bytes_aligned = ((dispatch_recv_count_buffer_bytes + 127) // 128) * 128  # align to 128
+        
+        # Calculate offset for buffer[0]
+        dispatch_recv_data_offset = signaling_buffer_bytes_aligned * 2 + send_buffer_bytes * 2
+        
+        # Set the offset for all proxies
+        for proxy in proxies:
+            proxy.set_dispatch_recv_data_offset(dispatch_recv_data_offset)
```
MaoZiming: @CalebZ9909 I think a good API is not setting `dispatch_recv_data_offset` directly, but pass in minimal information and proxy will do the above calculation itself (instead of doing this in the test script).

CalebZ9909:
fixed, I just moved the offset computation to uccl_proxy with an api. Working on atomic now, will commit here soon

MaoZiming:
@CalebZ9909 can you rebase from main?

CalebZ9909:
It seems Chihan is also editing the same files as me, so there are some conflict files that I'd better not solve right now. Maybe it is better to rebase until we both work, or we can rebase the same version of the combined code at this point?

CalebZ9909:
@MaoZiming Hi Ziming, I just pushed the code now. I tested the atomic-add on my side, and right now the atomic encoding/decoding and address translation look good. But the final write to dst gpu is not completed. Here are the changes:
1: [proxy.cpp](https://github.com/uccl-project/uccl/blob/a16c468e7248900533af9ab856e2d4040075adc9/gpu_driven/src/proxy.cpp#L351) add post_gpu_commands_mixed to handle both atomic and regular rdma, and it is called [here](https://github.com/uccl-project/uccl/blob/a16c468e7248900533af9ab856e2d4040075adc9/gpu_driven/src/proxy.cpp#L253). I used a 32 bits to represent the RDMA write with immediate, detailed explanation [here](https://github.com/uccl-project/uccl/blob/a16c468e7248900533af9ab856e2d4040075adc9/gpu_driven/src/proxy.cpp#L412)
2. [rdma.cpp](https://github.com/uccl-project/uccl/blob/a16c468e7248900533af9ab856e2d4040075adc9/gpu_driven/src/rdma.cpp#L697) decode here.
3. All offsets that GPU passes to CPU are with respect to dispatch_rdma_recv_data_buffer, and CPU_proxy adds the offset with respect to rdma_buffer to it to get the correct address. The encoding/decoding can support either a negative or positive offset, depending on whether the target write location is before or after dispatch_rdma_recv_data_buffer.

Now, I am working on the final write of atomic to update the counter on the remote GPU. Let me know if you think there is any issue with the above steps.

gpu_driven/src/rdma.cpp
```diff
@@ -675,6 +693,85 @@ void remote_process_completions(ProxyCtx& S, int idx, CopyRingBuffer& g_ring,
...
+          .wr_id = (0xa70a000000000000ULL | static_cast<uint64_t>(target_index & 0xFFFF)),  // Special marker for atomic ops
+          .dst_dev = 0,  // Assume device 0 for now
+          .src_ptr = reinterpret_cast<void*>(static_cast<uintptr_t>(atomic_value)),  // Store atomic value in src_ptr
+          .dst_ptr = reinterpret_cast<void*>(local_target_addr),  // Target GPU memory address
+          .bytes = sizeof(int)  // Size of atomic operation
+      };
+      
+      // Push to ring buffer for GPU processing (same as regular data transfer)
+      std::vector<CopyTask> atomic_tasks = {atomic_task};
+      while (!g_ring.pushN(atomic_tasks.data(), atomic_tasks.size())) {
```
MaoZiming: @CalebZ9909 this atomic task should be blocking. Here, because `g_ring` is handled by a separate thread, it might complete before the rdma write with immediate completes.

MaoZiming:
Otherwise LGTM!

CalebZ9909:
@MaoZiming @YangZhou1997 Hi Yang and Ziming, atomic part is ready for review now.

gpu_driven/src/peer_copy.cu
```diff
@@ -45,6 +45,38 @@ __global__ void peer_copy_kernel_vec(CopyTask const* __restrict__ tasks,
   if (task_id >= num_tasks) return;
 
   const CopyTask t = tasks[task_id];
+  
+  if (threadIdx.x == 0) {
```
MaoZiming: @CalebZ9909 Can these be removed?

MaoZiming:
@CalebZ9909 Remember to run `format.sh`

gpu_driven/src/peer_copy_worker.cpp
```diff
@@ -40,26 +40,46 @@ void peer_copy_worker(PeerCopyShared& shared, PeerWorkerCtx& ctx,
       }
       copy_batch_size = 1;
       ctx.tasks[0] = t;
+      
+      // Check for atomic tasks in single task mode
+      uint64_t wr_id = t.wr_id;
+      uint64_t masked = wr_id & 0xFFFF000000000000ULL;
```
MaoZiming: These parts can be removed. Only need to keep `IBV_WR_ATOMIC_FETCH_AND_ADD`

gpu_driven/src/peer_copy_worker.cpp
```diff
@@ -40,26 +40,46 @@ void peer_copy_worker(PeerCopyShared& shared, PeerWorkerCtx& ctx,
...
+        }
+      }
     }
     for (int i = 0; i < copy_batch_size; ++i) {
       maybe_enable_peer_access(shared.src_device, ctx.tasks[i].dst_dev);
       ctx.task_wrs[i] = ctx.tasks[i].wr_id;
     }
     ctx.highest_issued_wr_id =
         std::max(ctx.highest_issued_wr_id, ctx.task_wrs[copy_batch_size - 1]);
+
```
MaoZiming, line 97: Delete extra newlines

gpu_driven/src/proxy.cpp
```diff
@@ -98,11 +116,24 @@ void Proxy::run_dual() {
...
   uint64_t my_tail = 0;
   size_t seen = 0;
+  auto last_print = std::chrono::steady_clock::now();
   while (ctx_.progress_run.load(std::memory_order_acquire)) {
     poll_cq_dual(ctx_, finished_wrs_, finished_wrs_mutex_, cfg_.block_idx,
                  ring);
     notify_gpu_completion(my_tail);
     post_gpu_command(my_tail, seen);
+
+    auto now = std::chrono::steady_clock::now();
```
MaoZiming, line 126: If we comment out the printf, we can also remove the `last_print`, `now`, etc.

gpu_driven/src/proxy.cpp
```diff
@@ -169,7 +200,9 @@ void Proxy::post_gpu_command(uint64_t& my_tail, size_t& seen) {
   cmds_to_post.reserve(batch_size);
 
   for (size_t i = seen; i < cur_head; ++i) {
+
```
MaoZiming: Remove extra newlines (maybe format.sh will help you with this).

gpu_driven/src/proxy.cpp
```diff
@@ -293,6 +327,14 @@ void Proxy::run_local() {
       printf("Received command 1: block %d, seen=%d, value: %d\n",
              cfg_.block_idx + 1, seen, cmd_entry.value);
     }
+    // TransferCmd& cmd_entry = cfg_.rb->buf[idx];
```
MaoZiming: Remove.

gpu_driven/src/rdma.cpp
```diff
@@ -572,17 +585,29 @@ void local_process_completions(ProxyCtx& S,
...
-            S.largest_completed_wr = wr_done;
-            S.has_received_ack = true;
-            // printf("Local thread %d received ACK for WR %lu\n", thread_idx,
-            //        wr_done);
+
+          // Check if this is an atomic operation ACK (special WR ID format)
+          bool is_atomic_wr =
+              (wr_done & 0xFFFF000000000000ULL) == 0xa70a000000000000ULL;
+
+          if (is_atomic_wr) {
```
MaoZiming, line 591: Is this needed?

gpu_driven/src/rdma.cpp
```diff
@@ -678,7 +703,9 @@ void remote_process_completions(ProxyCtx& S, int idx, CopyRingBuffer& g_ring,
     if (wc[i].opcode != IBV_WC_RECV_RDMA_WITH_IMM) {
       continue;
     }
+
```
MaoZiming, line 704: Extra newlines

MaoZiming (approved):
After these code cleaning is done. Should be ready to go

gpu_driven/include/uccl_ibgda.cuh
```diff
@@ -65,19 +65,21 @@ __device__ __forceinline__ void nvshmemi_ibgda_put_nbi_warp(
...
       break;
     } else {
       auto now = clock64();
       if (now - last_print > 10 * 1e9) {
         uint64_t tail_cmd = rb->buf[cur_tail & rb->mask()].cmd;
-        printf(
-            "[ibgda_put_nbi_warp] %p waiting sm_id: %d, lane_id: %d, cur_head: "
-            "%llu, cur_tail: %llu, inflight: %llu, tail_cmd: %llu\n",
-            rb, sm_id, lane_id, cur_head, cur_tail, inflight, tail_cmd);
+        // printf(
```
MaoZiming: Remove the prints (if you decide to comment them out. including the `now`, etc.

gpu_driven/src/internode_ll.cu
```diff
@@ -271,6 +261,9 @@ __global__ __launch_bounds__(1024, 1) void dispatch(
   // Issue count sends
   if (responsible_expert_idx < num_experts and sub_warp_id == 0 and
       lane_id == 0) {
+    // printf("[ATOMIC_CHECK] About to execute atomic operation:
```
MaoZiming: Check if there are redundant comments.

gpu_driven/src/proxy.cpp
```diff
@@ -303,4 +345,141 @@ void Proxy::run_local() {
...
+    post_atomic_operations(atomic_wrs, atomic_cmds);
+  }
+}
+
+void Proxy::post_atomic_operations(
+    std::vector<uint64_t> const& wrs_to_post,
+    std::vector<TransferCmd> const& cmds_to_post) {
+  // Use RDMA hardware atomic operations directly on remote GPU memory
+
+  if (cmds_to_post.size() > ProxyCtx::kMaxAtomicOps) {
```
MaoZiming, line 371: `kMaxAtomicOps` where is this defined?

PR #302, opened 2025-08-26
Add IPC mem handle cache to P2P; fix UDS connection error

derekwin:


derekwin:
**Change Log:**

1.  **`p2p/engine.h`**: Added a 2D map `conn_id_and_ptr_to_ipc_cache_` (key: `conn_id` and data pointer, value: new `IpcCache` struct). This cache stores the remote peer's IPC memory handle for `send_ipc` and the locally created handle for `recv_ipc`.
2.  **`p2p/engine.cc`**: Implemented logic in `recv_ipc` and `send_ipc` to cache IPC memory handles.
3.  **`p2p/engine.cc`**: Released IPC memory handle resources during endpoint destruction.

derekwin:
**Performance Comparison vs. Commit `bca4eda` (AMD GPUs)**
current `bbedc9a`
Command: `torchrun --nnodes=1 --nproc_per_node=4 --node-rank=0 benchmark_uccl_collective.py --ring`

| Size   | Before (4 GPUs) | Before (6 GPUs) | This (4 GPUs) | This (6 GPUs) |
| :----- | :-------------- | :-------------- | :------------ | :------------ |
| 256B   | 0.02            | 0               | 0.06          | 0             |
| 1KB    | 0.09            | 0               | 0.24          | 0.01          |
| 4KB    | 0.36            | 0.02            | 0.96          | 0.03          |
| 16KB   | 0.18            | 0.08            | 3.84          | 0.13          |
| 64KB   | 5.57            | 0.59            | 13.64         | 1.19          |
| 256KB  | 21.52           | 1.83            | 50.1          | 5.14          |
| 1MB    | 74.35           | 8.67            | 145           | 13.32         |
| 10MB   | 251.09          | 8.74            | 307.46        | 24.35         |
| 16MB   | 287.3           | 20.1            | 332.68        | 25.32         |
| 100MB  | 364.92          | 74.57           | 379.07        | 80.62         |

(unit: Gbps)

derekwin:
Test failed with `--nproc_per_node=8`: https://github.com/uccl-project/uccl/issues/303

YangZhou1997:
I previously also found that when scaling to more than 4 GPUs, the performance drops severely. But not sure why

derekwin:
Although the UDS connection error has been resolved, two new issues now appear intermittently.
In particular, performance degrades when --nproc_per_node=8 is used, which may indicate inefficiencies in the current collective communication implementation. https://github.com/uccl-project/uccl/issues/303

p2p/engine.cc
```diff
@@ -1675,17 +1741,48 @@ bool Endpoint::recv_ipc(uint64_t conn_id, void* data, size_t size,
...
+  IpcCache ipc_cache;
+  bool found_in_cache = false;
+  {
+    std::shared_lock<std::shared_mutex> lock(ipc_cache_mu_);
+    auto it_conn = conn_id_and_ptr_to_ipc_cache_.find(conn_id);
+    if (it_conn != conn_id_and_ptr_to_ipc_cache_.end()) {
+      auto& ptr_to_ipc_cache = it_conn->second;
+      auto it_buf = ptr_to_ipc_cache.find(data);
+      if (it_buf != ptr_to_ipc_cache.end()) {
+        ipc_cache = it_buf->second;
```
YangZhou1997, line 1753: I think we cannot rely on `data` to decide if the sender-passed handler is the same, as in real applications, the receiver might supply the same `data` for different sender-passed handlers. 

Instead, we need to maintain some `ipc_id` by letting the application explicitly call reg_mem() and dereg_mem(), pass the `ipc_id` from the sender to the receiver, then the receiver looks up the cache based on `conn_id` and sender-passed `ipc_id`.

YangZhou1997 (commented):
It seems that you already merged the PR. But the current ipc_cache seems not correct, please see my comment. You can open a new PR to fix it.

p2p/engine.cc
derekwin: got it

PR #325, opened 2025-09-03
Fix GPU-NIC mapping initialization

jinqinn:
## Description
P2P engine attempted to initialize GPU-to-device mappings for all possible GPU indices (0-7), but get_best_dev_idx() only had mappings for actually detected GPUs.


YangZhou1997:
LGTM! Thank you @jinqinn

PR #336, opened 2025-09-05
Integrate intranode.cu to uccl

HermesCui:
## Description
1. Test the correctness of  `get_ipc_p2p_ptr` on one node with 8 gpus
2. Integrate intranode.cu to uccl gpu driven

### DIFF
--- a/ep/bench/buffer.py
+++ b/ep/bench/buffer.py
@@ -584,6 +584,265 @@ class Buffer:
         )
 
     # noinspection PyTypeChecker
+    def dispatch(
+        self,
+        x: Union[torch.Tensor, Tuple[torch.Tensor, torch.Tensor]],
+        handle: Optional[Tuple] = None,
+        num_tokens_per_rank: Optional[torch.Tensor] = None,
+        num_tokens_per_rdma_rank: Optional[torch.Tensor] = None,
+        is_token_in_rank: Optional[torch.Tensor] = None,
+        num_tokens_per_expert: Optional[torch.Tensor] = None,
+        topk_idx: Optional[torch.Tensor] = None,
+        topk_weights: Optional[torch.Tensor] = None,
+        expert_alignment: int = 1,
+        num_worst_tokens: int = 0,
+        config: Optional[Config] = None,
+        previous_event: Optional[EventOverlap] = None,
+        async_finish: bool = False,
+        allocate_on_comm_stream: bool = False,
+    ) -> Tuple[
+        Union[Tuple[torch.Tensor, torch.Tensor], torch.Tensor],
+        Optional[torch.Tensor],
+        Optional[torch.Tensor],
+        List[int],
+        Tuple,
+        EventOverlap,
+    ]:
+        """
+        Dispatch tokens to different ranks, both intranode and internode settings are supported.
+        Intranode kernels require all the ranks should be visible via NVLink.
+        Internode kernels require the ranks in a node should be visible via NVLink, while the ranks with the same GPU
+            index should be visible via RDMA.
+
+        Arguments:
+            x: `torch.Tensor` or tuple of `torch.Tensor`, for the first type, the shape must be `[num_tokens, hidden]`,
+                and type must be `torch.bfloat16`; for the second type, the first element of the tuple must be shaped as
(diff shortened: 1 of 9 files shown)


ep/bench/buffer.py
```diff
@@ -91,14 +91,55 @@ def __init__(
...
 
         # Synchronize IPC handles
         ipc_handles = [
             None,
         ] * self.group_size
         local_ipc_handle = self.runtime.get_local_ipc_handle()
+        
+        # CRITICAL: Ensure we're on correct device before IPC all_gather_object
+        torch.cuda.set_device(expected_device)
+        torch.set_default_device(f"cuda:{expected_device}")
```
MaoZiming: There are multiple duplicated `set_device` and `set_default_device`. Can clean

ep/include/common.hpp
```diff
@@ -26,7 +26,7 @@
 #define kNumThBlocks 6
 #define kNumThPerBlock 1
 #define kObjectSize 10752  // 10.5 KB
-#define kMaxOutstandingSends 2048
+#define kMaxOutstandingSends 64
```
MaoZiming: @HermesCui This is way too low for rdma. any reason for 64?

ep/include/proxy.hpp
```diff
@@ -43,18 +44,44 @@ class Proxy {
...
+      for (int d = 0; d < nDevices; ++d) {
+        ctx_.per_gpu_device_buf[d] = buf;
+      }
+    } else {
+      // Original behavior: allocate on all GPUs (for single GPU or unspecified scenarios)
+      for (int d = 0; d < nDevices; ++d) {
+        GPU_RT_CHECK(gpuSetDevice(d));
+        void* buf = nullptr;
+        GPU_RT_CHECK(gpuMalloc(&buf, total_size));
+        ctx_.per_gpu_device_buf[d] = buf;
```
MaoZiming: On main, `per_gpu_device_buf` is not used

ep/src/internode_ll.cu
```diff
@@ -159,7 +159,7 @@ __global__ __launch_bounds__(1024, 1) void dispatch(
             slot_idx * num_bytes_per_msg;
         auto const dst_p2p_ptr =
             ipc_base_ptrs ? uccl::get_ipc_p2p_ptr(dst_ptr, ipc_base_ptrs, rank,
-                                                  dst_rank, max_nvl_peers, 0)
+                                                  dst_rank, max_nvl_peers, 1000000000)
```
MaoZiming: The last field is `buffer_size`, where is this used in `get_ipc_p2p_ptr`?

ep/src/rdma.cpp
```diff
@@ -135,8 +135,45 @@ void per_thread_rdma_init(ProxyCtx& S, void* gpu_buf, size_t bytes, int rank,
...
-  // Assuming fixed GPU idx for now
-  int gpu_idx = 0;
+  // Determine correct GPU index based on rank for multi-GPU scenarios
+  int gpu_idx;
+  cudaGetDevice(&gpu_idx);
+  
+  // Check if we're in a multi-GPU single-node scenario
+  // In this case, each rank should use its corresponding GPU
+  int local_rank = -1;
+  const char* local_rank_env = getenv("LOCAL_RANK");
```
MaoZiming: In the future we can directly initialize UcclProxy with rank, because not all programs are launched with correct env variables set

ep/src/internode_ll.cu
HermesCui: Yeah, it is useless right now

ep/include/proxy.hpp
```diff
@@ -43,18 +44,44 @@ class Proxy {
   explicit Proxy(Config const& cfg) : cfg_(cfg) {
     const size_t total_size = kRemoteBufferSize;
     int nDevices;
-    cudaError_t err = cudaGetDeviceCount(&nDevices);
-    if (err != cudaSuccess) {
-      printf("CUDA error: %s\n", cudaGetErrorString(err));
-      std::abort();
+    cudaGetDeviceCount(&nDevices);
+    
+    // 保存当前设备
```
MaoZiming: Why do we need to preserve existing device id? 
Each proxy should be pinned to the correct GPU (so if the original device id is wrong, it should probably be changed somehow

PR #341, opened 2025-09-08
Fix IpcCache: 

derekwin:
## Description
[p2p | collective] Let the receiver maintain the IpcCache

Not ready yet, needs more consideration.

Fixes # (issue)
https://github.com/uccl-project/uccl/pull/302

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [x] Unit tests
- [x] Integration tests
- [x] Manual testing

## Checklist
- [ ] I have added tests.

### DIFF
--- a/p2p/engine.cc
+++ b/p2p/engine.cc
@@ -137,21 +137,23 @@ Endpoint::~Endpoint() {
   }
 
   // free IpcMemHandleCache
-  std::unique_lock<std::shared_mutex> lock(ipc_cache_mu_);
-  for (auto& conn_pair : conn_id_and_ptr_to_ipc_cache_) {
-    auto& ptr_map = conn_pair.second;
-    for (auto& buf_pair : ptr_map) {
-      IpcCache& ipc_buf = buf_pair.second;
-      if (ipc_buf.direct_ptr) {
-        if (ipc_buf.is_send) {
-          GPU_RT_CHECK(gpuIpcCloseMemHandle(ipc_buf.direct_ptr));
-        }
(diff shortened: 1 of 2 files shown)


YangZhou1997:
Hi @derekwin , I just got some ideas on how to enable transparent lifecycle management of RDMA memory registrations and IPC memory handler. The idea works like this: 

- Write a torch binding to hook the tensor allocation and deallocation actions. Eg, https://github.com/ovg-project/kvcached/blob/main/csrc/torch_bindings.cpp 
- As long as a tensor is not free, its virtual address is unique. Therefore, we can use the virtual address of the tensor as the identifier in a hash table to cache the mr_handler. 
- When the tensor is allocated, the hook will register it with RDMA.When the tensor is deallocated the hook will deregistarte it with RDMA
- For IPC memory handler lifecycle, on the importer GPU, we can use the tensor virtual address in the exporter GPU as the identifier in a hash table. This way we can avoid repeatedly opening an IPC memory handler. 
- When the exporter GPU deallocates the tensor, it will notify all the exporter GPU to close the IPC memory handler and remove the corresponding hash table entry.

derekwin:
good idea.
So, a wrapped tensor allocation interface will be provided. Tensors used for communication will be allocated through this new interface and will be automatically managed. Communication APIs only need to query the corresponding mr handle or ipc handle via virtual address to issue advise.

derekwin:
I will try it on this PR.

PR #367, opened 2025-09-12
Making ep work on gh200

MaoZiming:
## Description
* Make `test_low_latency.py` work on GH200

Fixes # (issue)
After a few successful runs: 
* Getting `AssertionError`

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual testing

## Checklist
- [ ] I have added tests.

### DIFF
--- a/ep/Makefile
+++ b/ep/Makefile
@@ -14,6 +14,24 @@ else
   EFA_LDFLAGS := -L$(EFA_HOME)/lib -lefa
 endif
 
+ARCH := $(shell uname -m)
+GPU_NAME := $(shell nvidia-smi --query-gpu=name --format=csv,noheader | head -n1)
+CPU_IS_ARM64 := 0
+GPU_IS_HOPPER := 0
+ifeq ($(ARCH),aarch64)
+  CPU_IS_ARM64 := 1
+endif
+ifneq (,$(findstring GH200,$(GPU_NAME)))
+  GPU_IS_HOPPER := 1
+endif
+ifeq ($(and $(CPU_IS_ARM64),$(GPU_IS_HOPPER)),1)
+  $(warning GH200 detected, building with GH200)
+  GH_CFLAGS := -DUSE_GRACE_HOPPER
+else
+  $(warning GH200 not detected, building without GH200)
+  GH_CFLAGS :=
+endif
+
 PYTHON            ?= python3
 PYBIND11_INCLUDES := $(shell $(PYTHON) -m pybind11 --includes)
 EXT_SUFFIX        := $(shell $(PYTHON) -c "import sysconfig as s; print(s.get_config_var('EXT_SUFFIX') or '.so')")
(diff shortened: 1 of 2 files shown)


YangZhou1997:
Now, it works pretty well now in terms of performance. The root cause is that `nvshmemi_ibgda_put_nbi_warp` needs to pass `dst_expert_local_idx` as the ring selector to align with the later `nvshmemi_ibgda_amo_nonfetch_add` that also uses `dst_expert_local_idx` as the ring selector. In this way, you could guarantee the writes and atomics to the same expert are using the same QP with order guarantee. 
image

The only thing is that the bench_kineto() sometimes fails: 
image

MaoZiming:
I see `nvshmemi_ibgda_amo_nonfetch_add` wrongly used `dst_expert_local_idx` instead of `warp_id` as the ring selector. should both use warp_id as the ring selector?

MaoZiming:
I am a little bit surprised about the performance

MaoZiming:
image
I am able to run on EFA, the performance is worse than what you posted above.

ep/src/internode_ll.cu
```diff
@@ -750,7 +750,7 @@ __global__ __launch_bounds__(1024, 1) void combine(
         nvshmemi_ibgda_put_nbi_warp(
             dst_ptr - reinterpret_cast<uint64_t>(rdma_buffer_ptr), buf_ptr,
             hidden * sizeof(nv_bfloat16), dst_rank,
-            warp_id,  // NOTE(MaoZiming): use warp_id for rb
+            local_expert_idx,  // NOTE(MaoZiming): use warp_id for rb
```
MaoZiming: We shall clean the comments and variable naming.

YangZhou1997:
looking at the ms-level latency of sending and high variance, it should hit some mutex sleep. So maybe try to remove the mutex for finished wrs? 

The latency of receive side looks great

YangZhou1997:
@MaoZiming , once merged with main, the performance gets sudden drop: 

image

Should we try to fix the performance bug in this PR or merge this first? 

Update: I just realized the lambda side is running some heavy tasks in the second node, which might be the root cause of low recv perf. I will test again later once that heavy task is done. 

Update 2: I think it is ready to merge this PR

MaoZiming:
@YangZhou1997 Should be good to merge. I was just debugging on this branch and found a weird issue that leads to correctness bugs. 
There seems to be some numerical instability with `kUseUE8M0` and `round_scale`. I fixed this by adding a small change in `per_token_cast_back` to 
```
x_scales = x_scales.view(dtype=torch.uint8).to(torch.int) << 23
```
instead of 
```
x_scales = x_scales.view(dtype=torch.int8).to(torch.int) << 23
```
The strange thing is that the original DeepEP code uses `torch.int8`.

To reproduce, try 
```
torchrun --nnodes=2 --nproc_per_node=1 --node_rank=0   --master_addr=10.1.227.34 --master_port=12355   bench/test_low_latency.py --num-tokens=512   --hidden=7168 --num-topk=4 --num-experts=64
```

YangZhou1997:
Interesting. I guess it might relate to num-tokens---DeepEP only tests 128 for their internode_ll. With 512, maybe it causes overflow of a uint8 thus some rounding bugs. Anyway, let me merge this PR now.

PR #373, opened 2025-09-14
Multi-Writer-Multi-Reader communication

CalebZ9909:
## Description

Fixes # (issue)

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [x] Unit tests
- [x] Integration tests
- [ ] Manual testing

## Checklist
- [ ] I have added tests.

### Tasks
- [x] Add Flat Combination
- [x] Write benchmark tests to test throughput and latency
- [ ] Align to Ring Buffer same API
- [x] Test across FC and other MWMR methods (lock ring buffer and evenly create ring buffers for all warps)
- [ ] Add batch sending later
- [x] Benchmark throughput on GH200
- [ ] Update documentation


MaoZiming:
Can it provide the same interface as `ring_buffer.cuh`?

CalebZ9909:
Yes, the most important ones are submit_request (warp writes to pub list), and run_combiner(combinator warp writes to ring buffer)

MaoZiming:
I mean providing the similar APIs as the existing `RingBuffer` so that it can be readily tested in the existing code. E.g. `volatile_store_cmd` and `atomic_set_and_commit`. 
```
template <typename T, FlowDirection Dir, uint32_t Capacity>
struct alignas(128) RingBuffer
```

CalebZ9909:
Sure, let me work on this, this might need some special parameters, I will double check

CalebZ9909:
Added the throughput test first, needs perf tuning. Will push the optimized one soon.
**Target: we want to hit 7Mops/s per GPU**

CalebZ9909:
After combinator scanning optimization, still only hit ~30% of the target tput goal. More optimazation needed

YangZhou1997:
@CalebZ9909 , can you paste the results or attach a screenshot of the tput and latency now achieved, as long as the brief cmd to run it, so we can have a rough sense of the performance numbers (and whether it is reasonable)

MaoZiming:
> After combinator scanning optimization, still only hit ~30% of the target tput goal. More optimazation needed

How many producers?

CalebZ9909:
@YangZhou1997 @MaoZiming 
run **make -f Makefile.fc clean && make -f Makefile.fc release**  and then **./test_fc_throughput**
For now, I am using 4 cpu proxy, and if I assign 64 warps, tput is 1.89 Mops/s,128 warps: 1.90 Mops/s, 256 warps: 2.06 Mops/s
I am optimizing the memory access now.

YangZhou1997:
Thank you for sharing it! I assume it is on p5en? We had some old results on H100+CX7 that reached 6Mops under 1:1 mapping between SMs and CPU cores. 

image

Maybe also try on GH200: I am curious if GH200 unified memory may have a better number.

CalebZ9909:
These results are from the GH200 node; I can also test on p5en. By the way, do you mean one CPU proxy handling one SM hits 6 Mops/s?

YangZhou1997:
The results I was showing: one the left, one proxy thread, on the right, proxy thread count = num of SMs.

CalebZ9909:
No-contention design can achieve >10Mops/s in this test bench:

**Warps:  32, Proxies:  4, Rings/proxy:  8, Throughput: 11.27 Mops/sec
Warps:  64, Proxies:  4, Rings/proxy: 16, Throughput: 21.39 Mops/sec
Warps: 128, Proxies:  4, Rings/proxy: 32, Throughput: 41.84 Mops/sec
Warps: 256, Proxies:  4, Rings/proxy: 64, Throughput: 83.30 Mops/sec**

 significantly outperforming flat combination and lock-ring-buffer designs. Will post results on p5en and current CAS perf comparison. The best design will be integrated into the codebase.

CalebZ9909:
Run "make all" to compile, then
1. **./test_multi_ring_throughput** for no-contention design
2. **./test_fc_throughput** for Flat Combination design
3.  **./test_lock_throughput** for lock-ring-buffer design

MaoZiming:
Interesting! I think >10Mops is definitely enough for our purpose since NIC might not able to match up with that.

CalebZ9909:
Yes, there might be some tput drop in DeepEP integration as I just wrote dummy content in the ring buffer for the tests. The real scenario should be much more complicated.

CalebZ9909:
Some weird bugs on p5en, will push code for that after fixing the bugs

YangZhou1997:
Keep a note of some performance numbers: 4 CPU proxies, 32 wraps. 

Flat Combination: 1.67 Mops/s
each warp one ring buffer: 6.52Mops/s
Ring buffer with lock design: 0.8Mops/s
current CAS design: 0.32Mops/s

PR #374, opened 2025-09-15
Debug why p5en two GPU internode ll only has 3GB/s bandwidth

YangZhou1997:
## Description

Fixes # (issue)

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual testing

## Checklist
- [ ] I have added tests.


YangZhou1997:
`ep/bench/test_low_latency.py` gives on p5en EFA

image

It achieved nearly full of 200G NIC (1 port of CX7) on GH200: 

image

Increasing the number of tokens from 128 to 256 does not help p5en performance. 

With standard perftest at 20 inflight 8KB messages, ib_write_bw can achieve 173Gb/s: 

```
ib_write_bw -d rdmap85s0 --report_gbits -x 0 -c SRD -t 20 -Q 1 -q 6 -s 8192 -F --use_cuda 0
ib_write_bw -d rdmap85s0 --report_gbits -x 0 -c SRD -t 20 -Q 1 -q 6 -s 8192 -F --use_cuda 0 10.1.227.34
```

ib_write_lat gives 11.61 x 2 = 23.22 us network RTT (it is only 2.6 us on GH200+CX7):
```
ib_send_lat -d rdmap85s0 -x 0 -c SRD -s 8192 -F --use_cuda 1
ib_send_lat -d rdmap85s0 -x 0 -c SRD -s 8192 -F --use_cuda 1 10.1.227.34
```

As a reference, Broadcom Thor2 gives 9.4 us RTT for same-rail, and 13.0 us for cross-rail. 

@zhongjiechen, any thoughts on this? cc @MaoZiming

YangZhou1997:
It is getting better when I avoid setting `efa_attr.sl = EFA_QP_LOW_LATENCY_SERVICE_LEVEL`

image

MaoZiming:
Let me try tomorrow.

YangZhou1997:
Some findings: 

When I try to print out how many GPU requests a proxy thread typically pulls once, I find EFA has: 
> Thread 1 post_gpu_commands_mixed: rdma_wrs.size()=1, atomic_wrs.size()=0

, while GH200+IB has: 
> Thread 1 post_gpu_commands_mixed: rdma_wrs.size()=20, atomic_wrs.size()=3

zhongjiechen:
I will take a look too when supporting dual NICs in p5en

YangZhou1997:
I just wrote a simple test that first sends a 1MB message, then sends an 8B message: it turns out that the 8B message first arrives at the server. So there is no ordering guarantee in EFA. cc: @MaoZiming @zhongjiechen

MaoZiming:
Thanks. I also made some changes (since EFA requires every wr to be signalled), the following selective signalling that we had before is not correct (though probably overwritten by efa so it doesn't pose an issue)
```
qpx->wr_flags = (j + 1 == k) ? IBV_SEND_SIGNALED : 0;
```

YangZhou1997:
Early notify GPU once local_poll_completion sees IBV_WC_RDMA_WRITE, getting slightly better perf

image

YangZhou1997:
I will run bench_remote.cu https://github.com/uccl-project/uccl/blob/main/ep/bench/benchmark_remote.cu to pinpoint the low-performance issue

YangZhou1997:
Okay, I think the low performance is because of the long latency of RDMA write in EFA, basically, the queueing delay on EFA NICs (as these software-based NICs are too slow to handle a high message rate). From the following profiling results when running benchmark_remote.py with a WRITE size of 10.5KB, we find that the network latency dominates the operation time. 

Looking at our previous 200-300 us dispatch and combine latency is already the optimal, as the main bottleneck is the EFA's long operation latency. cc @MaoZiming @zhongjiechen 

image

Commands to run: 
```
# Node 0
torchrun --nnodes=2 --nproc_per_node=1 --node_rank=0 \
  --master_addr=x.x.x.x --master_port=12356 \
  bench/benchmark_remote.py

# Node 1
torchrun --nnodes=2 --nproc_per_node=1 --node_rank=1 \
  --master_addr=x.x.x.x --master_port=12356 \
  bench/benchmark_remote.py
```

YangZhou1997:
The same benchmark results on GB200, with latency reduced by 90 us. cc @MaoZiming @zhongjiechen 

image

YangZhou1997:
With local_process_completions() IBV_WC_RDMA_WRITE signaling RDMA write competition, and IBV_SEND_SIGNALED for every WRITE, we get better results on p5en and GH200: 

image

image

ep/src/rdma.cpp
```diff
@@ -855,19 +841,21 @@ void local_process_completions(ProxyCtx& S,
                       wr_done, slot);
               std::abort();
             }
-            auto& vec = S.wr_id_to_wr_ids[wr_done];
```
MaoZiming, line 858: We need to clean `wr_id_to_wr_ids` since we are not relying on ack. This won't affect correctness. I will include that in
https://github.com/uccl-project/uccl/pull/380

MaoZiming (approved):

ep/src/rdma.cpp
YangZhou1997: Sounds good!

YangZhou1997:
Final update on the p5en we can get: 

image

PR #384, opened 2025-09-17
Intranode test ll

HermesCui:
## Description
Intranode Kernel Test 

Run with
```
OMP_NUM_THREADS=8 torchrun --standalone --nproc_per_node=8 bench/test_intranode.py \
    --num-tokens 2048 --hidden 3584 --num-topk 4 --num-experts 128
```


YangZhou1997:
@HermesCui @MaoZiming just let me know, once you have resolved the conflict and format. I can take a look with Chihan these days.

HermesCui:
Yeah, there are still 3 file `buffer.py`, `utils.py`, `proxy.cpp` that need to resolved.

MaoZiming:
taking a look now

ep/bench/utils.py
```diff
@@ -480,24 +475,21 @@ def initialize_uccl(scratch, scratch_nbytes, rank, num_ranks, group):
...
         proxy = ep.Proxy(
             rb_addr=bench.ring_addr(i),
             block_idx=i,
             gpu_buffer_addr=scratch_ptr,
             total_size=scratch_nbytes,
             rank=rank,
-            node_idx=node_idx,
-            local_rank=local_rank,
-            peer_ip=peer_ip,
+            peer_ip="" if is_intranode else peer_ip,  # 关键：intranode 时使用空字符串
```
MaoZiming: Avoid using chinese comments.

ep/src/proxy.cpp
```diff
@@ -59,7 +59,11 @@ void Proxy::init_common() {
   per_thread_rdma_init(ctx_, cfg_.gpu_buffer, cfg_.total_size, my_rank,
                        cfg_.block_idx, cfg_.local_rank);
   pin_thread();
-  if (!ctx_.cq) ctx_.cq = create_per_thread_cq(ctx_);
+
+  // Skip CQ creation if RDMA is not initialized (intranode mode)
+  if (ctx_.context && !ctx_.cq) {
```
MaoZiming: @HermesCui A good practice is to do 
```
if (rdma not initiated) {
   return; 
}
```
To avoid a lot of diff from the main

ep/bench/test_intranode.py
```diff
@@ -3,6 +3,11 @@
 import torch
```
MaoZiming, line 9: Can you add the example command (as the comment) at the top of the file?

ep/bench/test_intranode.py
```diff
@@ -44,36 +49,48 @@ def test_main(
...
+    )
+
+    x = torch.ones((num_tokens, hidden), dtype=torch.bfloat16, device=device) * rank
+
+    x_pure_rand = torch.randn((num_tokens, hidden), dtype=torch.bfloat16, device=device)
     x_e4m3 = per_token_cast_to_fp8(x) if Buffer.is_sm90_compiled() else None
     x_e4m3 = (x_e4m3[0], x_e4m3[1].T.contiguous().T) if x_e4m3 is not None else None
     scores = (
-        torch.randn((num_tokens, num_experts), dtype=torch.float32, device="cuda").abs()
+        torch.randn((num_tokens, num_experts), dtype=torch.float32, device=device).abs()
```
MaoZiming: `device=device` is this change needed if default device is set?

ep/src/proxy.cpp
```diff
@@ -98,79 +113,88 @@ void Proxy::init_common() {
...
-                             &remote_infos_[peer]);
-    if (remote_infos_[peer].addr != peers_[peer].ptr) {
-      fprintf(stderr,
-              "Rank %d block %d: Warning: remote addr mismatch for peer %d: "
-              "got 0x%lx, expected 0x%lx\n",
-              my_rank, cfg_.block_idx, peer, remote_infos_[peer].addr,
-              peers_[peer].ptr);
-      std::abort();
+  // Out-of-band exchange per pair (skip if RDMA not initialized)
+  if (ctx_.context) {
```
MaoZiming: Same here. Please use 
if (!ctx.context) continue; 
To avoid large diff

ep/tests/test_internode_ll.py
```diff
@@ -0,0 +1,596 @@
+#!/usr/bin/env python3
+"""
+Test for internode_ll kernels with support for both single-node and multi-node configurations.
```
MaoZiming: Is this test testing something different from `test_low_latency.py`?

MaoZiming:
@HermesCui I tried to merge and left some comments.

MaoZiming (changes requested):
Thanks!

ep/include/uccl_proxy.hpp
```diff
@@ -32,8 +32,10 @@ class UcclProxy {
 
   void* get_atomic_buffer_ptr() {
     if (!atomic_buffer_ptr_) {
-      fprintf(stderr, "Error: atomic_buffer_ptr_ is not set yet\n");
-      std::abort();
+      // For intranode mode, return a dummy pointer (atomic operations not
```
MaoZiming: Passing pointer like this is not safe. We can just avoid calling `get_atomic_buffer_ptr` in `test_intranode.py`?

ep/include/proxy.hpp
```diff
@@ -9,6 +9,7 @@
 #include <algorithm>
 #include <atomic>
 #include <chrono>
+#include <cstdlib>
```
MaoZiming: Where is this used?

ep/bench/utils.py
```diff
@@ -136,6 +136,8 @@ def get_cpu_proxies_meta(rank, scratch_ptr, scratch_bytes, num_ranks, group):
     all_meta = [None] * num_ranks
     device_index = int(os.environ.get("LOCAL_RANK", 0))
     torch.cuda.set_device(device_index)
+    # CRITICAL: Check device before all_gather_object in get_cpu_proxies_meta
```
MaoZiming: I think these devices check are no longer needed. Since we already do `set_device` before that

ep/bench/utils.py
```diff
@@ -455,20 +457,13 @@ def bench_kineto(
...
+    device_index = int(os.environ["LOCAL_RANK"])
 
-    local_rank = int(os.environ["LOCAL_RANK"])
-    nproc_per_node = int(os.environ.get("LOCAL_WORLD_SIZE", 1))
-    node_idx = rank // nproc_per_node
-
-    if int(os.environ.get("WORLD_SIZE")) % nproc_per_node != 0:
-        raise ValueError("WORLD_SIZE must be divisible by LOCAL_WORLD_SIZE")
+    peer_ip = get_peer_ip(rank, num_ranks, group)
+    is_intranode = peer_ip == "127.0.0.1" or peer_ip == ""
```
MaoZiming: We can pass in an argument `is_intranode` to `initialize_uccl` to indicate whether setting up uccl for intranode or internode.

ep/tests/test_internode_ll.py
MaoZiming: If this is similar, maybe we can remove this.

ep/src/proxy.cpp
```diff
@@ -98,79 +113,88 @@ void Proxy::init_common() {
     if (c.tag >= ctx_by_tag_.size()) ctx_by_tag_.resize(c.tag + 1, nullptr);
     ctx_by_tag_[c.tag] = &c;
 
-    c.context = ctx_.context;
-    c.pd = ctx_.pd;
-    c.mr = ctx_.mr;
-    c.rkey = ctx_.rkey;
+    // Only set RDMA resources if they were initialized
+    if (ctx_.context) {
```
MaoZiming: This can be avoided with a special flag `is_intranode`

ep/bench/test_intranode.py
MaoZiming: i just added

ep/src/uccl_proxy.cpp
```diff
@@ -7,6 +7,9 @@ UcclProxy::UcclProxy(uintptr_t rb_addr, int block_idx,
                      int node_idx, int local_rank, std::string const& peer_ip,
                      int num_experts, int num_ranks)
     : peer_ip_storage_{peer_ip}, thread_{}, mode_{Mode::None}, running_{false} {
+  if (peer_ip.empty()) {
+    printf("Intranode mode. UcclProxy returns\n");
```
MaoZiming, line 11: @HermesCui I just added this.

MaoZiming:
image

MaoZiming:
Using `ld_cg_global` might have some perf impact. 
See.  https://github.com/deepseek-ai/DeepEP?tab=readme-ov-file#notices

PR #420, opened 2025-10-02
try to merge redundant code in transport.cc

YangZhou1997:
## Description

Fixes # (issue)

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual testing

## Checklist
- [ ] I have added tests.


YangZhou1997:
@zhongjiechen do you remember how to fix this https://github.com/uccl-project/uccl/blob/92e8d54a5510eb1c648395538f6fffd5fef0c9ba/collective/rdma/transport.cc#L2573, where cqe_desc is not defined? It seems that cqe_desc cannot be derived from wr_id.

zhongjiechen:
> @zhongjiechen do you remember how to fix this
> 
> https://github.com/uccl-project/uccl/blob/92e8d54a5510eb1c648395538f6fffd5fef0c9ba/collective/rdma/transport.cc#L2573
> 
> , where cqe_desc is not defined? It seems that cqe_desc cannot be derived from wr_id.

The address of cqe_desc should be assigned to wr_id(cqe_desc acts as a wrapper of wr_id) when calling ibv_post_send.

YangZhou1997:
But it seems that, for this function, SubUcclFlow* was assigned to wr_id, see: 

https://github.com/uccl-project/uccl/blob/92e8d54a5510eb1c648395538f6fffd5fef0c9ba/collective/rdma/transport.cc#L2562-L2574

zhongjiechen:
> But it seems that, for this function, SubUcclFlow* was assigned to wr_id, see:
> 
> https://github.com/uccl-project/uccl/blob/92e8d54a5510eb1c648395538f6fffd5fef0c9ba/collective/rdma/transport.cc#L2562-L2574

Oh, I see. I think this part of the code can be removed, because the premise of measuring cc reaction delay is that we can get HW RX timestamp (using cq_ex instead of ordinary cq). Therefore, this code is invalid (it may have been copied by mistake when supporting NIC that only supports ordinary cq)

PR #431, opened 2025-10-09
Changes to support nixl test

praveingk:
## Description
NIXL tests create two NIXL agents and tests communication. I have added the following changes to support uccl engine tests within a single NIXL test process :
1) When two uccl engines are created by the same process, the second UCCL engine already sees rdma_ctl initialized with all devices populated. Hence, added a check (`rdma_ctl->num_devices ==0`) in `init_devs` to do rdma discovery. Otherwise, there are duplicate entries.
2) Observed that p2p_listen_fds[] might be 0, which is valid. Hence, initialized to -1, and modified the CHECK to `>=0` instead of `!=0`, since invalid fds are <0.
3) Added `inside_python` checks in engine.cc since nixl tests are launched from c++, and fail upon hitting pygil state checks. 
4) Added in_python while creating uccl_engine. This is true by default, but nixl tests will disable this, by passing a custom param.
5) Install `librdma_plugin` too, this is needed by the nixl test for linking. 

Reference of working NIXL test : https://github.com/praveingk/nixl-uccl/blob/uccl-plugin/test/gtest/plugins/uccl/uccl_test.cpp 
Fixes # (issue)

## Type of Change
- [ x] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ x] Unit tests
- [] Integration tests
- [x ] Manual testing

## Checklist
- [ ] I have added tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -94,3 +94,6 @@ ep/*.log
 nodes.txt
 pcie_test
 transport_test
+
+*.png
+ep/figs
--- a/ep/bench/test_internode.py
+++ b/ep/bench/test_internode.py
@@ -23,8 +23,6 @@ This benchmark verifies:
   * Top-k routing and per-expert token distribution
   * Compatibility with cached dispatch and low-latency kernels
   * Performance tuning for NVL and RDMA chunk sizes
-  
-It is currently still under development. 
 """
 
 import argparse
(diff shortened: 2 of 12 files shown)


praveingk:
@YangZhou1997 Thanks. Let me quickly test it out, and then merge.

PR #457, opened 2025-10-20
feat: EP port to amd

zhenhuang12:
## Description

Port ep code to amd, include intranode dispatch/combine, internode_ll dispatch
**fp8 low latency has precision issue**

Fixes # (issue)

## Type of Change
- [x] Bug fix
- [x] New feature
- [x] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing

## Checklist
- [ ] I have added tests.


MaoZiming:
@zhenhuang12 Great! will take a look soon

ep/src/uccl_proxy.cpp
```diff
@@ -15,10 +15,7 @@ UcclProxy::UcclProxy(int thread_idx, uintptr_t gpu_buffer_addr,
                      std::string const& peer_ip, int num_experts, int num_ranks,
                      int num_nodes)
     : peer_ip_{peer_ip}, thread_{}, mode_{Mode::None}, running_{false} {
-  if (peer_ip.empty()) {
```
zhenhuang12, line 18: EP 8 of internode_ll need atomic_buffer, directly return will cause illegal address access

YangZhou1997: Oh, yes. Nice observation!

YangZhou1997:
intranode works! 

```
[tuning] Best dispatch (BF16): SMs 64, NVL chunk 22, 174.84 GB/s (NVL), t: 280.88 us
[tuning] Best combine: SMs 64, NVL chunk 8: 107.59 GB/s (NVL), t: 456.43 us
```

When testing internode_ll, I hit this issue. I actually hit this previously, and it was caused by the mismatched Broadcom NIC driver on the container vs. on the host. One way is to import the host Broadcom driver into the container. Another is using wheels. 

image

@zhenhuang12, do you think it is possible to build a wheel inside the container, then outside of the container, we pip install the wheel? This would resolve the NIC driver mismatch issues for RDMA.

zhenhuang12:
> intranode works!
> 
> ```
> [tuning] Best dispatch (BF16): SMs 64, NVL chunk 22, 174.84 GB/s (NVL), t: 280.88 us
> [tuning] Best combine: SMs 64, NVL chunk 8: 107.59 GB/s (NVL), t: 456.43 us
> ```
> 
> When testing internode_ll, I hit this issue. I actually hit this previously, and it was caused by the mismatched Broadcom NIC driver on the container vs. on the host. One way is to import the host Broadcom driver into the container. Another is using wheels.
> 
> image
> @zhenhuang12, do you think it is possible to build a wheel inside the container, then outside of the container, we pip install the wheel? This would resolve the NIC driver mismatch issues for RDMA.

I think there has 3 ways to resolve this issue:
 1. build wheel inside the container and use it in the host
 2. mount host rdma lib to container
 3. rebuild driver in container
 
I prefer 1 and 2 because they are simple.

MaoZiming:
I hit 
```
[WARN] no NIC found, defaulting to empty
[FATAL] Selected RDMA NIC '' not found in verbs device list
```
When running `bench/test_intranode.py`, also caused by mismatched driver

YangZhou1997:
I am merging this pr. Now two issues remain, cc @MaoZiming : 

Hitting something for FIFO numaBind, but I checked that every input for it is either 0 or 1. 
image

Some correctness issues when running internode_ll, I guess this is now expected, as @zhenhuang12 mentions in the pr desc: 
> Port ep code to amd, include intranode dispatch/combine, internode_ll dispatch
> fp8 low latency has precision issue
image

Oh, ^ only happens to 1 nproc_per_node, with 8 nproc_per_node, I got 
image


Topology detection: I am adding an env variable like UCCL_IB_HCA, but our previous code should already support detecting the best NIC for each GPU based on pcie distance.

YangZhou1997:
Most of them are fixed in https://github.com/uccl-project/uccl/pull/470, except the errno 110. 
* It only appears with >= 2 nproc_per_node on our amd + broadcom testbed, but not on nebius h100 + cx7
* With 1 nproc_per_node, test_low_latency.py will hang or trigger a value error. 

Another note is that I find our AMD testbed is actually not cross-rail connected for all pairs of GPUs...., for example, rdma0 and rdma4 cannot connect on a certain pair of nodes

PR #469, opened 2025-10-24
fixing __CUDA_ARCH__ bugs

YangZhou1997:
## Description

Fixes # (issue)

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual testing

## Checklist
- [ ] I have added tests.

### DIFF
--- a/ep/src/internode.cu
+++ b/ep/src/internode.cu
@@ -88,7 +88,7 @@ __forceinline__ __device__ int translate_dst_rdma_rank(int const dst_rdma_rank,
   return kLowLatencyMode ? (dst_rdma_rank * NUM_MAX_NVL_PEERS + nvl_rank)
                          : dst_rdma_rank * NUM_MAX_NVL_PEERS;
 }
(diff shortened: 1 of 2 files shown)


YangZhou1997:
cc @zhenhuang12 @MaoZiming 

When NVCC compiles a .cu file, it compiles it twice:
Host compilation: `__CUDA_ARCH__` is undefined - this is for CPU code
Device compilation: `__CUDA_ARCH__` is defined with the target architecture (e.g., 800, 900) - this is for GPU code

So our `#if __CUDA_ARCH__` blocks are still compiled during host compilation, which means the code inside them is not available when the host code tries to use it. This causes internode.cu and internode_ll.cu errors. 

Solution is to use `__NVCC__` instead. These macros are defined whenever NVCC is compiling, regardless of whether it's host or device code. 

Note I tried `__CUDACC__` does not work, as it will be replaced by hipify to `__HIPCC__`.

PR #495, opened 2025-10-30
Add ucclEP wrapper

CalebZ9909:
## Description

Fixes # (issue)

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual testing

## Checklist
- [ ] I have added tests.

### DIFF
--- a/ep/uccl_ep/__init__.py
+++ b/ep/uccl_ep/__init__.py
@@ -1,0 +1,37 @@
+"""
+UCCL Expert Parallel (EP) Communication Library
+
+This package provides high-performance communication primitives for
+Mixture of Experts (MoE) models, supporting:
+- High-throughput intranode all-to-all (using NVLink)
+- High-throughput internode all-to-all (using RDMA and NVLink)
+- Low-latency all-to-all (using RDMA)
+"""
+
+import torch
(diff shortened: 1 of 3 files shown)


YangZhou1997:
Is this supposed to replace the bench/utils.py, bench/buffer.py? has the various test (test_intranode, test_low_latency, test_internode) worked with this ?

CalebZ9909:
Yes, we will have a specific folder for those but they are equivalent now, those in the bench are copy of that, future modifications will base on the ones in this new folder. I am still running the inter and intra tests. Also sglang integration side saw some issues when running the test and I will talk with them then do the rest of the completed wrap up work.

MaoZiming:
is there a way to avoid duplicating buffer.py and utils.py?

MaoZiming:
I think the sglang integration test should largely work now. See https://github.com/uccl-project/uccl/pull/496

CalebZ9909:
Yes, we can keep the original one or move to the new folder to put all the wrappers.

PR #521, opened 2025-11-06
[P2P] Full TCPX Plugin for NIXL

DanielDanyang:
## Description

Now it passed `python benchmarks/benchmark_nixl.py --backend tcpx --role server --sizes 67108864 --iters 1`
Performance: 4-7.5GB/s

Read and Multi-channel are not supported now.

### DIFF
--- a/collective/afxdp/misc/client_main.cc
+++ b/collective/afxdp/misc/client_main.cc
@@ -650,8 +650,8 @@ int main(int argc, char* argv[]) {
 
   client_addr_u32 = htonl(str_to_ip(client_ip_str));
   server_addr_u32 = htonl(str_to_ip(server_ip_str));
-  DCHECK(str_to_mac(client_mac_str, client_mac_char));
-  DCHECK(str_to_mac(server_mac_str, server_mac_char));
+  CHECK(str_to_mac(client_mac_str, client_mac_char));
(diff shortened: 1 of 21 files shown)


YangZhou1997:
Wow--This is big! great work @DanielDanyang @derekwin  The iter things should be some initialization bugs for the engine, I guess

p2p/uccl_engine.cc
```diff
@@ -100,13 +100,22 @@ void listener_thread_func(uccl_conn_t* conn) {
     switch (md.op) {
       case UCCL_READ: {
         tx_msg_t tx_data = md.data.tx_data;
+#ifdef USE_TCPX
```
derekwin: Since uccl_engine_reg already stores the MR information in mem_reg_info, the mr_id can be obtained just as in the original RDMA implementation. Hence, this code branch is redundant and can be removed.

MaoZiming:
A very small nit, I am wondering if we should use [P2P], etc. to label the PRs. similar to https://github.com/vllm-project/vllm/pulls and https://github.com/skypilot-org/skypilot/pulls @YangZhou1997

YangZhou1997:
Agree. We should conform to that. @DanielDanyang

PR #540, opened 2025-11-12
removing rkey!=0 check

YangZhou1997:
## Description

Fixes # (issue)

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual testing

## Checklist
- [ ] I have added tests.


YangZhou1997:
@MaoZiming I am removing the `rkey!=0` check, as we already check if the returned mr is null, which is the standard way to check if ibv_mr_reg fails or not. I think it is just that EFA NIC will occasionally return zero rkey. For example, I have seen `0xc1`, `0x4a` rkey values

PR #608, opened 2025-12-26
[Build] Fix Makefile header dependency tracking

### LINKED ISSUE #135: Makefile cannot sense file changes under include/
It was observed in `rdma`, not sure for other folders.

yurekami:
## Summary

Fix Makefiles to properly track header file dependencies, including headers in the `include/` directory.

**Problem:** The existing `DEPS = *.h` pattern only matches headers in the current directory, not in `$(ABS_REPO_ROOT)/include/`. This means changes to shared headers like `util/net.h` or `util/util.h` don't trigger rebuilds.

**Solution:** Use GCC's automatic dependency generation (`-MMD -MP` flags) which creates `.d` files containing the complete dependency chain for each source file.

## Changes

- `collective/rdma/Makefile` - Add automatic dependency generation
- `collective/rdma/Makefile.rocm` - Add automatic dependency generation  
- `collective/efa/Makefile` - Add automatic dependency generation

### Key modifications:
1. Add `-MMD -MP` flags to `CXXFLAGS`
2. Include generated `.d` files with `-include` directive
3. Remove static `DEPS = *.h` variable
4. Add `*.d` to clean target

## How it works

The `-MMD` flag tells GCC to generate a `.d` file for each compiled source, listing all headers it depends on. For example, `transport.d` might contain:
```makefile
transport.o: transport.cc transport.h ../../../include/util/net.h ../../../include/util/util.h
```

Now when any header changes, Make knows which objects need rebuilding.

## Test Plan

- [ ] Verify `make clean && make` works correctly
- [ ] Modify a header in `include/` and verify dependent objects rebuild
- [ ] Verify `.d` files are generated and cleaned up properly

Fixes #135

### DIFF
--- a/collective/efa/Makefile
+++ b/collective/efa/Makefile
@@ -8,8 +8,8 @@ EFA_HOME?=/opt/amazon/efa
 INC = -I./ -I$(CUDA_HOME)/include -I${EFA_HOME}/include -I$(ABS_REPO_ROOT)/include
 LIBS = -L ${CUDA_HOME}/lib64 -L ${EFA_HOME}/lib -libverbs -lefa -lcudart -lcuda -lpthread -lglog -lgflags -lgtest -lz -lelf
 LIBS_SHARED = ${LIBS}
-override CXXFLAGS += -O3 -g -std=c++17 -Wno-pointer-arith -Wno-interference-size -fPIC -DUSE_CUDA
-DEPS = *.h
+# Added -MMD -MP for automatic dependency generation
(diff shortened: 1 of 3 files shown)


YangZhou1997:
@zhongjiechen , what do you think of this? `-MMD -MP` sounds interesting?

zhongjiechen:
> @zhongjiechen , what do you think of this? `-MMD -MP` sounds interesting?

`.d` is a robust and easy-to-manage method, which I've used before, LGTM.

yurekami:
Thanks for the LGTM @zhongjiechen! Could a maintainer please merge this when ready?

PR #626, opened 2026-01-08
[EP] fix small error of ep build

monopodium:
## Description

Fixes # (issue)

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing

## Checklist
- [ ] I have added tests.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -273,7 +273,8 @@ echo "[2/3] Running build inside container..."
 DETECTED_GPU_ARCH=""
 if [[ "$BUILD_TYPE" =~ (ep|all) ]];then
   if [[ "$TARGET" == cuda* ]] && command -v nvidia-smi &> /dev/null; then
-    DETECTED_GPU_ARCH=$(nvidia-smi --query-gpu=compute_cap --format=csv,noheader 2>/dev/null | head -n1 | tr -d ' ')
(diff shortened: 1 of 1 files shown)


monopodium:
image

jq could not parse json with a warning string, and the build will fail after this jq error.

build.sh
```diff
@@ -273,7 +273,8 @@ echo "[2/3] Running build inside container..."
 DETECTED_GPU_ARCH=""
 if [[ "$BUILD_TYPE" =~ (ep|all) ]];then
   if [[ "$TARGET" == cuda* ]] && command -v nvidia-smi &> /dev/null; then
-    DETECTED_GPU_ARCH=$(nvidia-smi --query-gpu=compute_cap --format=csv,noheader 2>/dev/null | head -n1 | tr -d ' ')
+    DETECTED_GPU_ARCH="$(nvidia-smi --query-gpu=compute_cap --format=csv,noheader 2>/dev/null | head -n1 | tr -d ' ' || true)"
```
YangZhou1997, line 276: Do we need changes to CUDA? I think we only need it for amd right?

monopodium: This change just makes sure that the build script does not exit in Docker when this line cannot run successfully. Or do you just want the build script to fail when the gpu_arch cannot be detected?

YangZhou1997: Never mind, I think it looks good

Issue #605, opened 2025-12-25
[P2P] Potential Bug in Async Transfer


derekwin:
Currently, all async interfaces submit tasks as consumers to the following two queues:

```cpp
// For both net and ipc send/recv tasks.
jring_t* send_unified_task_ring_;
jring_t* recv_unified_task_ring_;
```

These tasks are then consumed by the corresponding two threads:

```cpp
std::thread send_proxy_thread_;
std::thread recv_proxy_thread_;
```

Taking `send_ipc` and `recv_ipc` as examples, a lock is only briefly held when querying the connection:

```cpp
// Get connection info
Conn* conn;
{
  std::shared_lock<std::shared_mutex> lock(conn_mu_);
  auto it = conn_id_to_conn_.find(conn_id);
  CHECK(it != conn_id_to_conn_.end())
      << "Connection not found for conn_id: " << conn_id;
  conn = it->second;
}
```

However, when accessing `conn->uds_sockfd_`, there is no corresponding mutual exclusion lock. As a result, when the send thread and the recv thread execute tasks on the same `conn` concurrently, they may simultaneously use the same UDS socket to transmit messages:

```cpp
// send_ipc
auto ret = uccl::receive_message_nonblock(
    sockfd,
    static_cast<void*>(&transfer_info),
    sizeof(transfer_info));

// recv_ipc
ret = uccl::receive_message_nonblock(
    client_fd,
    static_cast<void*>(&completion),
    sizeof(completion));
```

However, these two `uccl` message interfaces do not have a protocol to distinguish message boundaries or types, which may lead to message conflicts.

YangZhou1997:
Hi @derekwin can I bother you some time to help fix this issue?

derekwin:
No problem! Sorry I forgot after our last discussion.

PR #718, opened 2026-02-07
[P2P] Replaced UDS with Shm-JRing

derekwin:
## Description
- Replaced the original IPC mechanism with a shared-memory–based JRing implementation for local communication.
- All tests successfully passed using the following commands:
```
CUDA_VISIBLE_DEVICES=4,5,6 python tests/test_engine_send.py
CUDA_VISIBLE_DEVICES=4,5,6 python tests/test_engine_read.py
CUDA_VISIBLE_DEVICES=4,5,6 python tests/test_engine_write.py
CUDA_VISIBLE_DEVICES=4,5,6 python tests/test_engine_metadata.py
CUDA_VISIBLE_DEVICES=4,5,6 torchrun --nnodes=1 --nproc_per_node=2 tests/test_engine_nvlink.py

CUDA_VISIBLE_DEVICES=4,5,6 torchrun --nproc_per_node=2 benchmark_uccl.py --ipc

CUDA_VISIBLE_DEVICES=4,5,6 torchrun --standalone --nproc_per_node=3 benchmark_uccl_allgather.py --sizes 65536,262144,1048576 --num-iters 100
```

Fixes # ([605](https://github.com/uccl-project/uccl/issues/605))

## Type of Change
- [x] Bug fix
- [x] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [x] Unit tests
- [x] Integration tests
- [x] Manual testing

## Checklist
- [x] I have added tests.

### DIFF
--- a/p2p/tests/test_engine_nvlink.py
+++ b/p2p/tests/test_engine_nvlink.py
@@ -1,21 +1,11 @@
-#!/usr/bin/env python3
 """
-Test script for the UCCL P2P Engine
-using NVLink for inter-process communication.
+Test script for UCCL P2P Engine local NVLink IPC path.
 
 Run with:
-  OMP_NUM_THREADS=4 torchrun --nproc_per_node=2 test_engine_nvlink.py
-or:
-  python -m torch.distributed.run --nproc_per_node=2 test_engine_nvlink.py
+torchrun --nproc_per_node=2 test_engine_nvlink.py
 """
 
 import sys
-import os
-import time
-import struct
-import socket
-import numpy as np
-
 import torch
 import torch.distributed as dist
 
@@ -25,13 +15,10 @@ try:
     print("✓ Successfully imported p2p")
 except ImportError as e:
     print(f"✗ Failed to import p2p: {e}")
-    print("Make sure to run 'make' first to build the module")
     sys.exit(1)
 
 
-# parse_metadata is now provided by the C++ layer via p2p.Endpoint.parse_metadata()
-
-
+# Torch dist helper functions
 def _send_int(value: int, dst: int):
     t = torch.tensor([int(value)], dtype=torch.uint64)
     dist.send(t, dst=dst)
@@ -48,8 +35,7 @@ def _send_bytes(payload: bytes, dst: int):
     _send_int(n, dst)
     if n == 0:
         return
-    mv = memoryview(bytearray(payload))  # copies once, writable
-    buf = torch.frombuffer(mv, dtype=torch.uint8)  # no warning
+    buf = torch.frombuffer(memoryview(payload), dtype=torch.uint8)
     dist.send(buf, dst=dst)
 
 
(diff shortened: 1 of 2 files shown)


derekwin:
Shm-JRing indeed reduces complexity compared to UDS.
Although tests and runtime behavior appear correct, using a single JRing for communication among multiple concurrent processes may still lead to message confusion, similar to the issues previously observed with UDS ([#605](https://github.com/uccl-project/uccl/issues/605)).
One potential approach is to dedicate a separate thread per JRing (each GPU has exactly one JRing) to properly handle different types of messages originating from different GPUs.
Currently, we can buffer multiple ShmMsg entries within the ring, which might enable multiple processes to safely share a single queue for communication.
However, introducing per-JRing threads would increase overall system complexity.

YangZhou1997:
@derekwin Nice work! The shm-jring indeed reduces the system complexity a lot. 

I think only send and recv will use this shm-jring, and send and recv are assumed to be a pair on the two sides. Therefore, I feel there should not be any message confusion issues.

derekwin:
@YangZhou1997 
There are still some trade-offs to consider:
- Certain tests (e.g., tests/test_engine_nvlink) exhibit the following behavior: the client calls connect_local(), but the server does not call accept_local(). This works correctly with UDS and currently also works with the single shared-memory ring (where multiple processes on the same GPU share the same JRing). Both UDS and SHM-ring expose only a single file descriptor (FD) or handle. However, unlike socket(UDS),  SHM-ring lacks session semantics. As a result, if multiple processes on the same GPU concurrently use the same JRing for communication, message confusion or race conditions may occur.
- If we instead require a per-process SHM-ring, where each process pair uses its own dedicated ring pair, this would lead to a significant increase in the number of rings created. In this case, we’d need to assign a unique identifier (e.g., the process ID) when creating each ring. Consequently, the client would first need to discover the peer’s PID (e.g., via a temporary socket or another side channel) before it can attach to the correct ring. Additionally, the server would be required to call accept_local(). Should we retain the ability for the server to not call accept_local()?

YangZhou1997:
For tests/test_engine_nvlink, I feel only calling connect_local() should not work---that test probably gets broken due to some API semantic changes. I think we should assume connect_local() and accept_local() are called in paris. 

To be honest, I think per-process-pair SHM-ring might be the right solution. Dozens of connections in total in the worst case still look okay. We can use an encoding like `shm-rank-0-to-1` to avoid coordination overhead on the names. 

I have one additional question: what happens if a client tries to attach a ring that has not been created by the server (which might run late). Shall we have some fail-sleep-retry logic just like in the previous UDS?

derekwin:
Got it, I will update the test. Each endpoint will init its ingress JRing,(8 JRings for 8 GPUs), and connect_local() whill retry like UDS.

derekwin:
Hi @YangZhou1997 , this PR seems to be working as expected. (Unless many processes are launched on a single GPU, which, in practice, is unlikely in typical P2P use cases).

YangZhou1997:
Sounds good. I think the typical use case is one single process per GPU. Thank you @derekwin

YangZhou1997 (approved):
LGTM

PR #731, opened 2026-02-17
[EP] Add Intel RDMA NIC support for UCCL EP

manojgop:
Enable UCCL EP (Expert Parallelism) to work on Intel RDMA RoCEv2 NICs, which lack nvidia_peermem and have different RDMA capabilities than Mellanox ConnectX. All Intel-specific changes are gated behind the USE_INTEL_RDMA_NIC build flag, matching the pattern used in collective/rdma.

Build system:
- EP Makefile: USE_INTEL_RDMA_NIC=1 -> -DINTEL_RDMA_NIC compile flag
- EP setup.py: read USE_INTEL_RDMA_NIC env var -> -DINTEL_RDMA_NIC
- build.sh: pass USE_INTEL_RDMA_NIC to python3 setup.py build in build_ep()

DMA-BUF GPU memory registration (rdma.cpp, rdma.hpp):
- Add reg_mr_gpu_dmabuf() using cuMemGetHandleForAddressRange to obtain a DMA-BUF fd for GPU memory, then ibv_reg_dmabuf_mr to register it. This avoids the nvidia_peermem kernel module dependency.
- Handle PyTorch caching allocator suballocation by querying the real cudaMalloc base via cuMemGetAddressRange.
- Patch NULL mr->addr after ibv_reg_dmabuf_mr (irdma leaves it NULL, but downstream SGE calculations depend on it).
- Gate behind #ifdef USE_DMABUF with fallback to ibv_reg_mr_iova2.

Atomic buffer allocation (uccl_proxy.cpp):
- Use cudaHostAlloc (pinned host memory) instead of cudaMalloc for the atomic buffer, so ibv_reg_mr works on NICs without nvidia_peermem and CPU proxy threads can perform std::atomic operations on it.
- Gate behind #ifdef ATOMICS_USE_HOST_MEMORY.

Compile-time gating (common.hpp):
- #ifdef INTEL_RDMA_NIC enables USE_DMABUF and ATOMICS_USE_HOST_MEMORY.

HCA device detection (utils.py):
- Extend detect_ib_hca() to recognize irdma devices in addition to mlx5.

## Description

Fixes # (issue)

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual testing

## Checklist
- [ ] I have added tests.

### DIFF
--- a/ep/src/rdma.cpp
+++ b/ep/src/rdma.cpp
@@ -5,6 +5,10 @@
 #include "proxy_ctx.hpp"
 #include "rdma_util.hpp"
 #include "util/gpu_rt.h"
+#ifdef USE_DMABUF
+#include <cuda.h>
+#include <dlfcn.h>
+#endif
 #include <arpa/inet.h>
 #include <netinet/in.h>
 #include <algorithm>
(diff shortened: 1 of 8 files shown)


ep/src/rdma.cpp
```diff
@@ -38,6 +42,124 @@
...
+            "[RDMA] cuMemGetHandleForAddressRange failed (CUresult=%d), "
+            "falling back to ibv_reg_mr_iova2 (needs nvidia_peermem)\n",
+            (int)cu_err);
+    return ibv_reg_mr_iova2(pd, gpu_buf, bytes, iova, access);
+  }
+
+  // Register only the sub-range we care about via offset into the DMA-BUF fd.
+  ibv_mr* mr =
+      ibv_reg_dmabuf_mr(pd, offset_in_alloc, bytes, iova, dmabuf_fd, access);
+  if (!mr) {
```
MaoZiming, line 135: Will this nvidia_peermem fallback always fail now?

ep/src/uccl_proxy.cpp
```diff
@@ -63,8 +63,15 @@ UcclProxy::UcclProxy(int thread_idx, uintptr_t gpu_buffer_addr,
 #elif defined(EFA)
     cudaHostAlloc(&atomic_buffer_ptr_, kAtomicBufferSize,
                   cudaHostAllocMapped | cudaHostAllocWriteCombined);
+#else
+#ifdef ATOMICS_USE_HOST_MEMORY
+    // Pinned host memory: ibv_reg_mr works on NICs without nvidia_peermem
+    // (e.g. irdma), CPU proxy threads can do std::atomic ops, and GPU
+    // kernels can still access it via CUDA host-mapped addressing.
+    cudaHostAlloc(&atomic_buffer_ptr_, kAtomicBufferSize, cudaHostAllocMapped);
```
MaoZiming, line 71: We might eventually combine this with the EFA path as both requires cudaHostMalloc for atomics; We can postpone this later.

MaoZiming (approved):
LGTM

ep/src/rdma.cpp
manojgop: nvidia_peermem will not be available for Intel NIC since nvidia_peermem is incompatible with irdma driver. So nvidia_peermem will fail for Intel RDMA NIC. For non Intel RDMA NIC which wants to use DMABUF it'll fallback to ibv_reg_mr_iova2 ((needs nvidia_peermem) if for some reason DMABUF registration fails

ep/src/uccl_proxy.cpp
manojgop: ok

PR #737, opened 2026-02-19
[EP] Share RDMA context/PD/MR across proxy threads on same NIC

manojgop:
When multiple proxy threads select the same NIC for the same GPU buffer
    (common on single-NIC systems), each thread was opening a separate
    ibv_context and registering the entire GPU buffer via DMA-BUF. Under
    full IOMMU translation (intel_iommu=on without iommu=pt), each
    registration requires per-page IOMMU mappings for the full DMA-BUF
    scatter-gather list. With 4 threads each registering a 64 GB buffer,
    this exhausts IOMMU mapping resources, causing ENOMEM on
    ibv_reg_dmabuf_mr. (With iommu=pt, identity-mapped DMA bypasses
    per-page translation and this issue does not occur.)

    Add a per-NIC shared resource cache keyed by (nic_name, gpu_buf_ptr).
    The first thread to init creates the ibv_context / PD / MR normally and
    stores them in the cache. Subsequent threads on the same NIC reuse the
    cached resources. A refcount ensures the last thread to destroy actually
    frees the resources; earlier threads have their pointers nulled so
    cleanup is skipped.

    This reduces DMA-BUF registrations from num_threads to num_unique_NICs
    (4x -> 1x on single-NIC setups). Multi-NIC systems (EFA) are unaffected
    since threads selecting different NICs get separate resources. The cache
    is gated behind #ifdef USE_DMABUF.

    Note: this alone does not solve the IOMMU mapping limit for very large
    buffers — it only eliminates redundant registrations. The actual fix for
    registering buffers beyond the IOMMU limit is the chunked DMA-BUF
    registration in the follow-up commit.

## Description

Fixes # (issue)

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual testing

## Checklist
- [ ] I have added tests.

### DIFF
--- a/p2p/README.md
+++ b/p2p/README.md
@@ -766,6 +766,10 @@ python tests/test_engine_write.py
 python tests/test_engine_metadata.py
 torchrun --nnodes=1 --nproc_per_node=2 tests/test_engine_nvlink.py
 
+# One-sided IPC correctness tests (write_ipc, read_ipc, writev_ipc, readv_ipc — sync and async)
+# Verifies that each API correctly copies data from source to destination buffers.
+torchrun --nnodes=1 --nproc_per_node=2 tests/test_engine_onesided_ipc.py
+
 # One-sided IPC benchmarks (write_ipc / read_ipc)
 torchrun --nproc_per_node=2 benchmarks/benchmark_uccl.py --write-ipc
 torchrun --nproc_per_node=2 benchmarks/benchmark_uccl.py --read-ipc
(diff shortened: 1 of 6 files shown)


ep/src/rdma.cpp
```diff
@@ -338,9 +395,65 @@ void per_thread_rdma_init(ProxyCtx& S, void* gpu_buf, size_t bytes, int rank,
...
+    }
+    g_shared_rdma_cv.notify_all();
+  };
+#endif
+
   S.context = ibv_open_device(dev_list[selected_dev_idx]);
   if (!S.context) {
     perror("Failed to open device");
+#ifdef USE_DMABUF
+    signal_failure();
```
MaoZiming, line 455: Is `signal_failure()` needed since it is followed by exit(1)

manojgop: It's not needed today for correctness (the process dies anyway), but it's a safety net against future refactors. It makes intent explicit — a reader immediately understands that other threads depend on this entry

PR #743, opened 2026-02-22
[EP] Add benchmarking option for FP8 dispatch.

laochonlam:
Enable benchmarking with either FP8 dispatch or direct BF16 dispatch through a runtime flag for apples-to-apples performance comparison.

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual testing

## Checklist
- [ ] I have added tests.

### DIFF
--- a/ep/bench/test_low_latency.py
+++ b/ep/bench/test_low_latency.py
@@ -155,7 +155,9 @@ def test_main(
         for return_recv_hook in (False, True):
             for dispatch_use_fp8_case in (False, True):
                 for round_scale in (False,):
-                    for round_scale in (False, True) if dispatch_use_fp8_case else (False,):
+                    for round_scale in (
+                        (False, True) if dispatch_use_fp8_case else (False,)
+                    ):
                         for use_ue8m0 in (False, True) if round_scale else (False,):
                             print(
                                 "Start experiment with settings:"


ep/bench/test_low_latency.py
```diff
@@ -152,14 +153,14 @@ def _record_hash(label: str, t: torch.Tensor, include_in_overall: bool = True):
 
     for current_x in x_list:
         for return_recv_hook in (False, True):
-            for dispatch_use_fp8 in (False, True):
+            for dispatch_use_fp8_case in (False, True):
```
MaoZiming, line 156: Why do we have both `dispatch_use_fp8` and `dispatch_use_fp8_case`?

laochonlam: `dispatch_use_fp8_case` is already in the original correctness sweep, and it is used only to cover both FP8 and BF16 correctness paths.

The update here is benchmark-side control: `dispatch_use_fp8` is a runtime CLI option for selecting FP8 vs BF16 dispatch during measurement (originally only fp8=True)

The semantic for correctness loop and benchmark loop is a bit different, In order not to have naming confusion, we keep `dispatch_use_fp8_case` for correctness sweep coverage and use `dispatch_use_fp8` only as the benchmark runtime flag.

MaoZiming: Got it.

PR #759, opened 2026-02-26
Fix bug in EFA benchmarks BW calculation

yehuday:
Network BW in GBps is measured in decimal (10^9) and not binary (2^30). This drops ~7% from the actual bandwidth result.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing

## Checklist
- [ ] I have added tests.

### DIFF
--- a/.github/workflows/uccl-build-test.yml
+++ b/.github/workflows/uccl-build-test.yml
@@ -93,3 +93,169 @@ jobs:
         with:
           name: build-log
           path: build.log
+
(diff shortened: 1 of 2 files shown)


MaoZiming:
@yehuday Thanks! Good catch

PR #761, opened 2026-02-26
[EP]: chunked DMA-BUF MR registration for IOMMU mapping limits.

manojgop:
Under full IOMMU translation (intel_iommu=on without iommu=pt), each
    registration requires per-page IOMMU mappings for the full DMA-BUF
    scatter-gather list. For larger buffers, this exhausts IOMMU mapping
    resources, causing ENOMEM on ibv_reg_dmabuf_mr. (With iommu=pt,
    identity-mapped DMA bypasses per-page translation and this issue does
    not occur.)

    When the GPU buffer exceeds 2 GiB (e.g. ~63 GiB for num-tokens=4096),
    split it into 1 GiB chunks, each with its own DMA-BUF fd and MR.
    If the single-MR path succeeds, chunking never triggers.

    Key changes:
    - common.hpp: add kMaxDmabufChunkSize (1 GiB) and kMaxMRChunks (128)
    - proxy_ctx.hpp: add MRChunk/RemoteMRChunk/WRSegment structs, gpu_mr_chunks and remote_mr_chunks vectors, lkey_for()/rkey_for() address-based lookup, and split_for_chunks() which splits a single transfer into sub-WRs when it straddles a local or remote MR chunk boundary (each sub-WR uses the correct per-chunk lkey/rkey)
    - rdma.hpp: extend RDMAConnectionInfo with per-chunk metadata array, declare reg_mr_gpu_dmabuf_chunked()
    - rdma.cpp: implement reg_mr_gpu_dmabuf_chunked() with per-chunk fd export; fix reg_mr_gpu_dmabuf() and chunked variant to always use 2 MiB-aligned export sizes (cuMemGetAddressRange reports the requested size, not the actual rounded-up size, causing unaligned exports that cuMemGetHandleForAddressRange rejects); fall back to chunked on single-MR failure; call split_for_chunks() at all RDMA write posting sites (normal mode, software-ordering mode, and fast mode) so that any transfer crossing a chunk boundary is split into correctly-keyed sub-WRs before being posted to the QP
    - proxy.cpp: propagate chunks to peer contexts, store remote chunks from exchanged connection info, deregister all chunks on cleanup

    Most RDMA writes (~14 KB per MoE token) fit within a single chunk,
    but transfers whose address range spans two chunks (e.g. a write
    starting near a 1 GiB boundary) are transparently split by
    split_for_chunks() into multiple work requests.

## Description

Fixes # (issue)

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual testing

## Checklist
- [ ] I have added tests.

### DIFF
--- a/.github/workflows/uccl-build-test.yml
+++ b/.github/workflows/uccl-build-test.yml
@@ -71,13 +71,13 @@ jobs:
             fi
 
             # Create env if missing
-            conda info --envs | awk \x27{print \$1}\x27 | grep -qx uccl || conda create -y -n uccl python=3.11
+            conda info --envs | awk \x27{print \$1}\x27 | grep -qx uccl || conda create -y -n uccl python=3.12
             conda activate uccl
 
             cd /home/skytestuser/uccl-test
-            ./build.sh cuda all 3.11 --install 2>&1 | tee build.log
+            ./build.sh cuda all 3.12 --install 2>&1 | tee build.log
 
-            grep -q \"Successfully installed uccl-0.0.1.post4\" build.log
+            grep -q \"Successfully installed uccl-\" build.log
           "'
 
       - name: Fetch remote build.log
--- a/README.md
+++ b/README.md
@@ -144,8 +144,8 @@ source ~/miniconda3/bin/activate
 source ~/.bashrc # or .zshrc and others
 conda init
 
-# Install python ssh lib
-pip install paramiko pybind11
+# Install python ssh lib and more
+pip install paramiko intervaltree pybind11 nanobind
 # Upgrade conda glic to modern ones
 conda install -c conda-forge "libstdcxx-ng>=12" "libgcc-ng>=12"
 ```
--- a/build.sh
+++ b/build.sh
@@ -246,7 +246,7 @@ ${CONTAINER_ENGINE} "${CONTAINER_RUN_ARGS[@]}" \
   -e TORCH_CUDA_ARCH_LIST="${TORCH_CUDA_ARCH_LIST:-}" \
   -e DISABLE_AGGRESSIVE_ATOMIC="${DISABLE_AGGRESSIVE_ATOMIC:-0}" \
   -e UCCL_WHEEL_PLAT="${UCCL_WHEEL_PLAT:-}" \
-  -e FUNCTION_DEF="$(declare -f build_rccl_nccl_header build_ccl_rdma build_ccl_efa build_p2p build_ep build_ukernel)" \
+  -e FUNCTION_DEF="$(declare -f rename_to_abi3 build_rccl_nccl_header build_ccl_rdma build_ccl_efa build_p2p build_ep build_ukernel)" \
   -w /io \
   "$IMAGE_NAME" /bin/bash -c '
     set -euo pipefail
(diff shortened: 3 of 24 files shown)


ep/src/rdma.cpp
```diff
@@ -1427,17 +1634,23 @@ static void post_rdma_async_batched_normal_mode(
                 AtomicsImm::Pack(true, false, cmd.atomic_val, cmd.atomic_offset,
                                  get_low_latency(cmd.cmd_type))
                     .GetImmData();
-            wrs[j].opcode = IBV_WR_RDMA_WRITE_WITH_IMM;
-            wrs[j].imm_data = htonl(imm);
+            wrs.back().opcode = IBV_WR_RDMA_WRITE_WITH_IMM;
```
MaoZiming, line 1637: Does that mean for a write that spans two segments, it will get translated into two writes, with only an immediate for the second write?
Could the second write arrive first before the first write?

ep/include/common.hpp
```diff
@@ -45,6 +45,17 @@ extern bool use_ll_sl;
 #define ETHERNET_RDMA
 #endif
 
+#ifdef USE_DMABUF
+// Maximum size for a single DMA-BUF MR registration.  Under full IOMMU
+// translation (intel_iommu=on without iommu=pt), each registration requires
+// per-page IOMMU mappings for the full DMA-BUF scatter-gather list.  For
+// larger buffers, this exhausts IOMMU mapping resources, causing ENOMEM on
+// ibv_reg_dmabuf_mr if size is greater than 2 GiB.  Splitting into chunks of
```
MaoZiming, line 53: Is the 2GB NIC specific or the same for all DMABUF

manojgop: It's not NIC Specific. It's related to IOMMU. Issue comes from iommu that it's hard to get contiguous IOVA for 2+ GiB memory request. The ENOMEM comes from the kernel IOMMU subsystem running out of mapping resources, and any RDMA driver going through dma_buf_map_attachment() under full IOMMU translation would hit the same issue. We won't see the issue with iommu=pt (but gives weaker DMA isolation and also not suitable for VM).

ep/src/rdma.cpp
manojgop: Good question. For a write spanning two chunks:

1. Two RDMA WRITEs are posted: The first is a plain IBV_WR_RDMA_WRITE (no signal, no IMM). The second gets IBV_SEND_SIGNALED and, if applicable, IBV_WR_RDMA_WRITE_WITH_IMM.

2. No, the second cannot arrive before the first on the same QP. The IB/RDMA spec guarantees in-order delivery for work requests posted on the same QP. Since both sub-WRs are chained and posted together via a single ibv_post_send() call on the same QP , the RNIC must complete them in posting order. The first write's data is visible at the remote side before the second write's data lands.

This is why the code's approach is safe: the IMM (which triggers the atomic update on the receiver) is on the last sub-WR, so by the time the receiver processes the IMM completion, all preceding data from the first sub-WR is guaranteed to be in remote memory

ep/include/common.hpp
MaoZiming: Interesting that peermem doesn't require this. Thanks for the explanation!

PR #764, opened 2026-02-27
[EP] renaming uccl.ep to uccl_ep

YangZhou1997:
## Description

Fixes # (issue)

## Type of Change
- [ ] Bug fix
- [x] New feature
- [x] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual testing

## Checklist
- [ ] I have added tests.


YangZhou1997:
This PR separate uccl_ep as a independent wheel (rather than uccl.ep) with its own pyproject.toml and build.sh, suitable for future wheel release via pypi

cc @zhenhuang12 , @praveingk

MaoZiming:
@YangZhou1997 Could you remind me what do you mean by "uccl_ep as a independent wheel "? 
are we thinking about an interface like `from uccl import ep`?

YangZhou1997:
It is actually directly `import uccl_ep` instead of ep as part of uccl wheel, so ep and other parts can evolve independently.

MaoZiming:
Shouldn't the entire uccl have one release?

YangZhou1997:
I remember @praveingk once thinks better to have separate wheels?

praveingk:
@YangZhou1997 I am not opposed to having the same wheel for ep and p2p . I think the separation was intended more for collectives. In fact, same wheel might be better since just one installation is required for inference builds.

MaoZiming:
@YangZhou1997 @praveingk 
I think having the same wheel version for the entire repo is more clear (many open-sourced projects have multiple libraries, but they have a single wheel). But I agree that there might be some coordination (e.g. making sure all libraries are stable before releasing a wheel)

YangZhou1997:
Sounds good. Let's stick to a single wheel then!

PR #766, opened 2026-02-27
[EP] Fix off13 overflow and PER_EXPERT_BATCHING receiver barrier stride

whn09:
Fix three bugs that caused crashes and hangs when running low-latency EP benchmarks with `PER_EXPERT_BATCHING` on 32+ GPUs (4 nodes × 8 GPUs):

1. **off13 overflow in RDMA atomic immediate data**: The 13-bit `off13` field in `AtomicsImm` (max value 8191) was assigned raw byte offsets from `cmd.req_rptr` (up to 81960 bytes). With `PER_EXPERT_BATCHING` and 32 ranks, double-buffered offsets exceed 8191. Fix: encode as `req_rptr >> 3` on sender (4 locations), decode as `GetOff() << 3` on receiver (2 locations). This works because `req_rptr` always points into an `int64_t` array (8-byte aligned).

2. **Receiver barrier double-buffer stride mismatch**: In `remote_process_completions_fast_mode` (`USE_RECEIVER_BARRIER` path), the double-buffer stride was `align(num_experts * sizeof(int64_t), 128)`, but with `PER_EXPERT_BATCHING` the dispatch buffer uses `num_ranks * num_ranks * sizeof(int64_t)`. For 32 ranks / 9 local experts: computed stride=128 bytes vs correct 8192 bytes, causing wrong offset decomposition and infinite hang. Fix: use `max(num_ranks*num_ranks, num_experts)` matching `ep_config.hpp`.

3. **cudaMalloc(0) in buffer.py**: `num_rdma_bytes=0` causes `cudaMalloc` to return NULL which `torch::from_blob` rejects. Guard with `max(num_rdma_bytes, 1)`.

## Description

These three bugs are triggered together when using `PER_EXPERT_BATCHING` with ≥32 GPUs across multiple nodes. Bug 1 (off13 overflow) silently corrupts atomic offsets, bug 2 (stride mismatch) causes infinite hangs in the receiver barrier, and bug 3 (cudaMalloc(0)) crashes on certain configurations where RDMA buffer size computes to zero.

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing

Tested on 4× p5.48xlarge (32× H100 80GB, 32× EFA 100Gb/s):
- All correctness tests pass (BF16/FP8, sync/async, with/without top-k, with/without `return_recv_hook`)
- Intranode test (1 node, 8 GPUs): all 24 tests passed
- Low-latency EP32 benchmark (4 nodes, 32 GPUs, 256 experts): 4.16 GB/s D+C bandwidth (~5297 us avg)
- Normal mode EP32 benchmark (4 nodes, 32 GPUs, 288 experts): dispatch FP8 21.5 GB/s, BF16 39.4 GB/s
- `PER_EXPERT_BATCHING` shows ~20% improvement over non-batched baseline (3.49 → 4.16 GB/s)

## Checklist
- [ ] I have added tests.


MaoZiming:
@laochonlam let's add assert for byte-alignment

laochonlam:
Hi @whn09 , could you provide more information so we can reproduce the issue, such as the command you ran and the GPU/NIC environment? We tried to reproduce the results, but did not observe any errors on H100 4×8 machines on the current main (9a81ba3d738ce64a33b58b7b43b3923bc1ae1bb0). Thanks.

whn09:
## Reproduction Results: upstream main vs this PR on 4× p5.48xlarge (32 GPUs)

We ran the same low-latency benchmark on **upstream main** (`ea9b12bb`) and **this PR branch** with `PER_EXPERT_BATCHING=1` to verify the bug fix.

### Test Configuration

- **Hardware**: 4× p5.48xlarge (32× H100 80GB, 32× EFA 100Gb/s per node)
- **Software**: PyTorch 2.9.1+cu130, CUDA 13.0
- **Benchmark**: `bench/test_low_latency.py --num-tokens=128 --hidden=7168 --num-topk=8 --num-experts=256 --dispatch-use-fp8`
- **Build flag**: `PER_EXPERT_BATCHING=1`

### Results

| Version | Commit | PER_EXPERT_BATCHING | Result |
|---------|--------|---------------------|--------|
| **This PR** | `f1562c0e` | ✅ Enabled | ✅ All correctness tests pass. ~4.29 GB/s, ~5140 μs avg D+C latency |
| **Upstream main** | `ea9b12bb` | ✅ Enabled | ❌ **SIGABRT on all 32 ranks** (NCCL collective timeout → crash) |

### Why the bug only reproduces at 32+ GPUs

The two bugs in upstream main are scale-dependent:

1. **off13 overflow**: The 13-bit `off13` field in `AtomicsImm` (max value 8191) stores raw byte offsets from `cmd.req_rptr`. With `PER_EXPERT_BATCHING` + 32 ranks, `dispatch_recv_count_buffer = 32 × 32 × sizeof(int64_t) = 8192 bytes`. Double-buffered offsets exceed the 13-bit limit → silent data corruption → hang.

2. **Receiver barrier stride mismatch**: The double-buffer stride is hardcoded as `align(num_experts × sizeof(int64_t), 128)`. With 256 experts / 32 ranks = 8 local experts, this computes to only 128 bytes. But with `PER_EXPERT_BATCHING`, the actual buffer dimension is `num_ranks × num_ranks × sizeof(int64_t) = 8192 bytes`. Wrong stride → wrong `new_offset` → wrong `expert_idx/src_rank` decomposition → `is_atomic_ready` never becomes true → infinite hang → NCCL timeout.

At smaller scales (e.g., 2 nodes / 16 GPUs), `off13` values stay within range and the stride mismatch is less severe, which explains why this issue was not reproducible in smaller test environments.

whn09:
## Follow-up: upstream main WITHOUT PER_EXPERT_BATCHING runs fine

To further isolate the bug, we rebuilt upstream main (`ea9b12bb`) **without** `PER_EXPERT_BATCHING` and ran the same benchmark on the same 4× p5.48xlarge cluster.

### Updated Results

| Version | Commit | PER_EXPERT_BATCHING | Result |
|---------|--------|---------------------|--------|
| Upstream main | `ea9b12bb` | ❌ Off | ✅ All tests pass. 4.05 GB/s, 5448 μs avg |
| Upstream main | `ea9b12bb` | ✅ On | ❌ **SIGABRT on all 32 ranks** |
| This PR | `f1562c0e` | ✅ On | ✅ All tests pass. 4.29 GB/s, 5140 μs avg |

This confirms the crash is specifically in the `PER_EXPERT_BATCHING` code path, which this PR fixes. The ~6% bandwidth improvement (4.05 → 4.29 GB/s) also shows PER_EXPERT_BATCHING working as intended once the bugs are resolved.

laochonlam:
LGTM. Thanks!

PR #776, opened 2026-03-01
[Build] fix auditwheel repair failure on hosts with older glibc

manojgop:
By default, the wheel now keeps the container's glibc platform tag
(detected dynamically inside the container). On hosts with a different
glibc (e.g. Rocky Linux 9.4 with glibc 2.34), pip install fails:

ERROR: uccl-0.0.1.post4-cp38-abi3-manylinux_2_35_x86_64.whl is not a
supported wheel on this platform.

Setting UCCL_WHEEL_ENABLE_FORCE_RETAG=1 retags the wheel to the host's
glibc version, which resolves the install error. A warning is printed
when versions differ since the binaries are still linked against the
container's glibc and may use symbols absent in the host's older glibc.

## Description

Fixes # (issue)

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual testing

## Checklist
- [ ] I have added tests.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -13,6 +13,13 @@
 #                          Example: CONTAINER_ENGINE=podman ./build.sh cuda all
 #   USE_INTEL_RDMA_NIC=1   Enable Intel RDMA NIC support (irdma driver, vendor 0x8086)
 #                          Example: USE_INTEL_RDMA_NIC=1 ./build.sh cuda ccl_efa
+#   UCCL_WHEEL_ENABLE_FORCE_RETAG=1  Allow retagging the wheel to the host's
+#                          glibc version when it differs from the container's.
+#                          By default the wheel keeps the container's glibc tag.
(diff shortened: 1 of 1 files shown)


YangZhou1997:
Hi @manojgop , I think this is too risky as the host glibc might not be compatible with the container glibc. I would suggest adding an env variable allowing forcibly retagging the wheel to a specific glibc version. What do you think of?

manojgop:
> Hi @manojgop , I think this is too risky as the host glibc might not be compatible with the container glibc. I would suggest adding an env variable allowing forcibly retagging the wheel to a specific glibc version. What do you think of?

Added env variable UCCL_WHEEL_ENABLE_FORCE_RETAG to allow force retag the wheel

PR #778, opened 2026-03-02
[p2p] Add compression for send/recv

monopodium:
## Description

Fixes # (issue)

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [x] Integration tests
- [ ] Manual testing

## Checklist
- [ ] I have added tests.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -13,6 +13,13 @@
 #                          Example: CONTAINER_ENGINE=podman ./build.sh cuda all
 #   USE_INTEL_RDMA_NIC=1   Enable Intel RDMA NIC support (irdma driver, vendor 0x8086)
 #                          Example: USE_INTEL_RDMA_NIC=1 ./build.sh cuda ccl_efa
+#   UCCL_WHEEL_ENABLE_FORCE_RETAG=1  Allow retagging the wheel to the host's
+#                          glibc version when it differs from the container's.
+#                          By default the wheel keeps the container's glibc tag.
+#                          WARNING: the wheel is still built against the
+#                          container's glibc and may use symbols not present in
+#                          an older host glibc.
+#                          Example: UCCL_WHEEL_ENABLE_FORCE_RETAG=1 ./build.sh cuda all
 #
 # The wheels are written to wheelhouse-[cuda|rocm|therock]
 # -----------------------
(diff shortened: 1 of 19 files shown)


monopodium:
@YangZhou1997 This version does not support GB10 yet. But without USE_DIETGPU=1, nothing will be affected.

build.sh
```diff
@@ -422,4 +422,4 @@ if [[ "$DO_INSTALL" == "1" ]]; then
   else
     echo "Warning: Could not detect UCCL installation path"
   fi
-fi
+fi
```
YangZhou1997: avoiding modifying this file? leaving a empty line seems a standard manner.

p2p/rdma/compression.h
```diff
@@ -0,0 +1,593 @@
...
+#include "memory_allocator.h"
+
+/**
+ * @brief Abstract interface for compressor backends.
+ *
+ * This pure virtual interface defines the compression/decompression operations
+ * without any DietGPU-specific types. Implementations can use DietGPU or
+ * provide a null/stub implementation.
+ */
+class ICompressorBackend {
```
YangZhou1997, line 12: What does this "I" before CompressorBackend mean?

p2p/Makefile.rocm
YangZhou1997, line 1: Make Makefile also work with compression?

p2p/rdma/compression.h
monopodium: "I" means interface, an abstract class.

praveingk:
Very interesting! How is the performance with compression?

monopodium:
@YangZhou1997 Updated.

monopodium:
> Very interesting! How is the performance with compression?

It depends on which kind of GPU you use. On MI300X with Broadcom BCM57608 400G NIC, the throughput with two different compression strategies is the following:
image

YangZhou1997:
@monopodium Thanks for making the changes! Great work. Just one quick question: does the current DietGPU part ever use any torch C++ ABIs (eg, `#include <torch.h>`) or python C++ ABIs (eg, `#include <Python.h>`)? I am asking, as we have now made our P2P code compatible with any versions of Python by using Nanobind (pybind11 is not ABI stable), and not directly use torch c++ ABIs.

monopodium:
> @monopodium Thanks for making the changes! Great work. Just one quick question: does the current DietGPU part ever use any torch C++ ABIs (eg, `#include <torch.h>`) or python C++ ABIs (eg, `#include <Python.h>`)? I am asking, as we have now made our P2P code compatible with any versions of Python by using Nanobind (pybind11 is not ABI stable), and not directly use torch c++ ABIs.

DietGPU uses it, but the part integrated to P2P does not use it. One problem is that I will use setuptool to generate .hip files (I copy this method from ep, and I do not know how to turn .cu to .hip separately, and GPT also seems not to know; I will ask it again). In this file, https://github.com/uccl-project/uccl/blob/main/thirdparty/dietgpu/setup.py, it needs torch.utils.cpp_extensions.

PR #786, opened 2026-03-04
[P2P] Add APIs for Ray integration

zhongjiechen:
## Description

1. Add two new APIs: `deregister_memory` and `remove_remote_endpoint`
2. Support CPU memory in `(de)register_memory` APIs.
3. A temporary solution for intra-node CPU-CPU communication with `transfer()` API (RDMA loopback)

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual testing

## Checklist
- [ ] I have added tests.

### DIFF
--- a/p2p/engine_api.cc
+++ b/p2p/engine_api.cc
@@ -473,7 +473,7 @@ NB_MODULE(p2p, m) {
                     nb::cast<XferDesc const&>(local_desc_list[i]);
                 auto const& rdesc =
                     nb::cast<XferDesc const&>(remote_desc_list[i]);
-                CHECK(!rdesc.ipc_info.empty())
+                UCCL_CHECK(!rdesc.ipc_info.empty())
                     << "Remote descriptor has no IPC info for local transfer";
                 IpcTransferInfo info;
                 std::memcpy(&info, rdesc.ipc_info.data(), sizeof(info));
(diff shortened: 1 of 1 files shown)


zhongjiechen:
Now, Ray APIs should work on the following intra-node transfer:

1. [RDMA loopback] Intra-node CPU-CPU transfer.
2. [GPU IPC]             Intra-node GPU-GPU transfer.
3. [RDMA loopback]Intra-node GPU-CPU transfer.

3 will fail without GPUDirect RDMA.

Used commands:
```
torchrun --nnodes=1 --nproc_per_node=2 --master_addr=localhost benchmarks/benchmark_ray_p2p.py --intra --sender-device cpu/gpu --receiver-device cpu/gpu --perf
```

TODOs:
1. We can't handle testbeds without GPUDirect RDMA, and even the error log is unclear. (Another PR)
2. I use RDMA loopback to support intra-node CPU-CPU and GPU-CPU transfer for simplicity. Should we use other methods? e.g., async memcpy?

CC @YangZhou1997

YangZhou1997:
Thanks @zhongjiechen ! (Would be great if you could fix the conflict)

1. I think we can ignore no GPUDirect RDMA for now. As long as we can print out logs/warnings, we should be good. 
2. We actually have an intra-node GPU-CPU transfer using cudaMemcpy: basically, these APIs https://github.com/uccl-project/uccl/blob/main/p2p/engine.h#L407-L432 support host memory as well. But does Ray require CPU-CPU transfer?

zhongjiechen:
I see @YangZhou1997 

Jerry reported a CPU-CPU transfer issue, so I added a fix. I might have overlooked existing host memory support—let me check if I need to revert anything. I’ll resolve the conflicts and merge this PR quickly.

Issue #799, opened 2026-03-07
EP on AMD+Broadcom hit CQE 12 again


YangZhou1997:
The amd+broadcom cluster is kind of crazy, hitting CQE12 at the beginning of LL and hT. Feeling there are some config issues. cc @MaoZiming @laochonlam 

```
CQE ERROR wr_id=3 status=12(transport retry counter exceeded) opcode=1 byte_len=0 vendor_err=0xa qp_num=0x2c10
CQE ERROR wr_id=4294967296 status=12(transport retry counter exceeded) opcode=1 byte_len=0 vendor_err=0xa qp_num=0x2c14
CQE ERROR wr_id=4294967316 status=12(transport retry counter exceeded) opcode=1 byte_len=0 vendor_err=0xa qp_num=0x2c04
CQE ERROR wr_id=8589934593 status=12(transport retry counter exceeded) opcode=1 byte_len=0 vendor_err=0xa qp_num=0x2c0b
CQE ERROR wr_id=4294967296 status=12(transport retry counter exceeded) opcode=1 byte_len=0 vendor_err=0xa qp_num=0x2c03
CQE ERROR wr_id=4294967304 status=12(transport retry counter exceeded) opcode=1 byte_len=0 vendor_err=0xa qp_num=0x2c02
CQE ERROR wr_id=8589934592 status=12(transport retry counter exceeded) opcode=1 byte_len=0 vendor_err=0xa qp_num=0x2c04
CQE ERROR wr_id=4294967296 status=12(transport retry counter exceeded) opcode=1 byte_len=0 vendor_err=0xa qp_num=0x2c06
CQE ERROR wr_id=12 status=12(transport retry counter exceeded) opcode=1 byte_len=0 vendor_err=0xa qp_num=0x2c08
CQE ERROR wr_id=8589934594 status=12(transport retry counter exceeded) opcode=1 byte_len=0 vendor_err=0xa qp_num=0x2c01
CQE ERROR wr_id=8 status=12(transport retry counter exceeded) opcode=1 byte_len=0 vendor_err=0xa qp_num=0x2c09
CQE ERROR wr_id=1 status=12(transport retry counter exceeded) opcode=4 byte_len=0 vendor_err=0xa qp_num=0x2c08
```

laochonlam:
I am curious about this. I can take a look

YangZhou1997:
[...]

PR #803, opened 2026-03-09
[EP] Fix RoCE GID auto-detection for Broadcom bnxt_re NICs

laochonlam:
Found by manually setting UCCL_IB_GID_INDEX=3 which resolved the CQE errors on the Broadcom cluster. Solving #799.

## Reason: 
The existing GID auto-detection logic in ncclIbGetGidIndex defaults to AF_INET (IPv4-mapped) filtering. On Broadcom bnxt_re NICs (e.g. Thor-2) where all GIDs are IPv6 (link-local fe80:: and ULA fd93::), the filter skips every candidate and falls back to GID index 0, which is an unroutable link-local address. This causes IBV_WC_RETRY_EXC_ERR (status=12, vendor_err=0xa) on all RDMA operations.

## Fix:
Add a fallback that re-scans the GID table without the address-family constraint when the initially selected GID is invalid, preferring a valid RoCE v2 entry. Also log the auto-detected GID index to aid debugging.

Fixes # (issue)

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual testing

## Checklist
- [ ] I have added tests.

### DIFF
--- a/ep/include/rdma_util.hpp
+++ b/ep/include/rdma_util.hpp
@@ -351,6 +351,37 @@ static bool ncclIbGetGidIndex(struct ibv_context* context, uint8_t portNum,
     }
   }
 
+  // Fallback: when the selected GID is link-local or unconfigured (e.g. on
+  // Broadcom bnxt_re where all GIDs are IPv6 ULA, not IPv4-mapped), the
+  // address-family filter above skips every candidate.  Retry without the
+  // address-family constraint, preferring valid + RoCE v2.
+  {
+    union ibv_gid selectedGid;
+    if (ibv_query_gid(context, portNum, *gidIndex, &selectedGid) == 0 &&
+        !validGid(&selectedGid)) {
+      char const* deviceName = ibv_get_device_name(context->device);
+      int bestIdx = -1;
+      for (int i = 0; i < gidTblLen; ++i) {
+        union ibv_gid gid;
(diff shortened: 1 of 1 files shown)


ep/include/rdma_util.hpp
```diff
@@ -351,6 +351,37 @@ static bool ncclIbGetGidIndex(struct ibv_context* context, uint8_t portNum,
...
   }
 
+  // Fallback: when the selected GID is link-local or unconfigured (e.g. on
+  // Broadcom bnxt_re where all GIDs are IPv6 ULA, not IPv4-mapped), the
+  // address-family filter above skips every candidate.  Retry without the
+  // address-family constraint, preferring valid + RoCE v2.
+  {
+    union ibv_gid selectedGid;
+    if (ibv_query_gid(context, portNum, *gidIndex, &selectedGid) == 0 &&
+        !validGid(&selectedGid)) {
```
MaoZiming, line 361: Does this work on other platforms?

laochonlam: Let me test with gb10 soon. I have only tested it for amd+broadcom NIC

MaoZiming:
@laochonlam This seems to work on Intel

PR #815, opened 2026-03-13
fp8 support

monopodium:
## Description

Fixes # (issue)

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing

## Checklist
- [x] I have added tests.

### DIFF
--- a/p2p/benchmarks/benchmark_uccl_compression.py
+++ b/p2p/benchmarks/benchmark_uccl_compression.py
@@ -19,7 +19,6 @@ import torch.distributed as dist
 
 from uccl import collective
 
-
 _DTYPE_ITEM_SIZE = {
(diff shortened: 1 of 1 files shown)


monopodium:
@YangZhou1997 I just came up with a sample idea to deal with fp8, which is not perfect but can also gain some benefit. I just treat two FP8s as a unit and treat them in a similar way to BF16. One interesting finding is that the FP8 compression ratio of float compression and raw byte raw has almost no difference; I guess it is because the bits of FP8 are fully utilized to express a number. But float compression can achieve a little higher compression bandwidth.  image

PR #879, opened 2026-04-14
[P2P] Multi-transport Build and Dispatcher

praveingk:
## Description
Added an all-transport build, which builds all transport libraries, and uses a simple dispatcher to select the transport to link.
This will avoid compiling again for different transport selection. Useful for docker images where we don't need to choose transport option during build.

Currently this is a separate build option

At runtime:

`UCCL_P2P_TRANSPORT=rdma/tcp/tcpx/efa` can be set to choose the transport option.

Fixes # (issue)

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual testing

## Checklist
- [ ] I have run `format.sh` to follow the style guidelines.
- [ ] I have run `build.sh` to verify compilation.
- [ ] I have removed redundant variables and comments.
- [ ] I have updated the documentation.
- [ ] I have added tests.

@YangZhou1997 : Let me know if this is a good option to have. This is currently simple with a one-time dispatcher during launch time to do transport selection.

@derekwin @DanielDanyang For now, I have fixed the TCP-X to avoid using the IPC path in engine.cc, so that build is fine.

PR #881, opened 2026-04-15
[P2P] Merge TCPX onto the NCCL endpoint and fix the NCCL/TCPX runtime path

DanielDanyang:
## Description
These changes remove the old dedicated TCPX endpoint, move TCPX onto the shared NCCL endpoint, and fix the merged runtime path.

## Main changes

- remove the legacy `nccl_tcpx_endpoint` implementation and use the shared NCCL endpoint for the TCPX path
- fix NCCL/TCPX notification routing so notifications use the NCCL flow id instead of the outer UCCL connection id
- bypass the generic vector proxy path for single-iov NIXL transfers on the NCCL/TCPX backend
- use a single NCCL communicator per peer in `p2p/nccl/nccl_endpoint.cc`, which matches the working TCPX send/recv pattern
- disable the background passive-accept helper for the merged NCCL path to avoid the 8-GPU connection race
- document the working NCCL version and runtime env for the current hosts

## Validation

- 1-GPU `benchmark_nixl.py --backend uccl --op-type read` runs with `GPUDirectTCPX_v7`
- 8-GPU `benchmark_nixl_8gpu.py` runs successfully with all workers using `GPUDirectTCPX_v7`


praveingk (approved):
Thanks @derekwin . I tried it out, and seems to be working 👍🏽 
Once this is merged, I will rebase #879 on top of this.

derekwin:
LGTM!

derekwin (approved):

PR #884, opened 2026-04-15
[EP] Update latest vLLM LL results and gracefully handle num_tokens >= 256

MaoZiming:
## Description
Fixes an LL-dispatch hang on 2× p5en nodes when DeepSeek-V3-sized configs push per-WR num_tokens ≥ 256 (e.g. NT=4096, NEXP=128).

Root cause: `get_low_latency(cmd.cmd_type)` returns bit 7 of `cmd_type`, which encodes the **low-latency buffer index** (0/1), not an "is-LL" flag. The receiver side of `post_rdma_async_batched_fast_mode` was gating on it as if it were "is-LL"; when `buffer_idx=0` it fell into the non-LL decode path and read `num_tokens` from only the 8-bit `atomic_val`, silently truncating any value ≥ 256. The counter on the receiver never matched the dispatcher, and dispatch hung.

This path is LL-only, so the fix uses the LL decoders unconditionally: unpack 10-bit `expert_idx` + 13-bit `num_tokens` from the packed `expert_idx` slot (high 5 bits) plus `atomic_val` (low 8 bits) at the three sites in `post_rdma_async_batched_fast_mode`.

```
Bit layout (bits [31:30] reserved = 0 for write type):
  [29]    IS_COMBINE   (1 bit)
  [28]    BUFFER_IDX   (1 bit)
  [27:19] EXPERT_IDX   (9 bits, 0..511)
  [18:6]  NUM_TOKENS   (13 bits, 0..8191)
  [5:0]   RANK         (6 bits, 0..63)
```

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
- [ ] Unit tests
- [x] Integration tests
- [x] Manual testing

Reproduced the UCCL-EP Low Latency numbers from `ep/bench/vllm/README.md` end-to-end on 2× p5en with DeepSeek-V3-0324, `deepep_low_latency` backend, dp=2 tp=8, `vllm bench serve` at 10 RPS × 1000 prompts × 256 concurrency:

| Metric | README target | This branch (warm JIT) |
|---|---|---|
| Request throughput | 9.24 req/s | **9.28** |
| Output tok/s | 2365.44 | **2375.37** |
| Mean TTFT | 271.31 ms | **268.76** |
| P99 TTFT | 688.79 ms | **604.01** |
| Mean TPOT | 61.02 ms | **60.29** |
| P99 TPOT | 77.52 ms | **72.28** |

## Checklist
- [ ] I have added tests.


MaoZiming:
@YangZhou1997 I have verified that both LL and HT for vLLM work on 2 nodes of p5en.  I am not sure why we didn't hit this issue before.

YangZhou1997:
Super cool, Maybe we should update the vllm results in bench/vllm?

YangZhou1997 (approved):

PR #892, opened 2026-04-20
[EP]: fix async event chaining and auto-detect intranode topology

zhenhuang12:
## Description
Two related fixes for the EP bench `Buffer` APIs:

1. **Preserve tensor liveness across async dispatch/combine chains.** The normal-mode intranode/internode dispatch and combine paths now propagate `previous_event.extra_tensors` and append the current call's inputs / outputs / `handle` into the returned `EventOverlap` whenever `async_finish=True`. Previously these paths returned `EventOverlap(event)` with no recorded tensors, so a chained `dispatch → combine` could drop dependencies and race with the caching allocator.
2. **auto-detect intranode vs. internode topology from the `ProcessGroup`.** A new `detect_group_topology()` in `ep/bench/utils.py` uses `NODE_RANK` / `GROUP_RANK` / `SLURM_NODEID` / `socket.gethostname()` (first hit wins) as a per-node token and `all_gather_object`s it over the group, returning `(local_rank, node_idx, num_nodes, is_intranode)`. `initialize_uccl(..., is_intranode=None)` and `Buffer(..., is_intranode=None)` now default to this auto-detection and no longer rely on `LOCAL_WORLD_SIZE` / `WORLD_SIZE` env vars. Passing `is_intranode=True` on a multi-node group is now a hard `ValueError` instead of silently mis-configuring proxies.

Also moves the `allocate_on_comm_stream ⇒ (previous_event, async_finish)` precondition from C++ back to Python, because Python now owns the actual tensor allocation stream before handing raw pointers into C++; the six affected `EP_HOST_ASSERT` sites in `ep/src/uccl_ep.cc` are replaced with `static_cast<void>(allocate_on_comm_stream)`.

Fixes # (issue)

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing

## Checklist
- [ ] I have added tests.


YangZhou1997:
Let me @MaoZiming for reviewing. All CI passes.

YangZhou1997 (approved):
LGTM

PR #898, opened 2026-04-22
[EP] Handle zero-token intranode dispatch/combine paths

Matt-Harvill:
## Description
Fix intranode dispatch/combine for cases where a rank logically receives 0 tokens.

This change keeps kernel pointer arguments valid in zero-recv cases by allocating a minimal backing buffer and then slicing returned tensors back to the logical size. It also adds a regression test that forces one rank to receive no tokens and verifies dispatch, cached dispatch, and combine still behave correctly.

Fixes # N/A

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 

- Added a regression test in `ep/bench/test_intranode.py` for a zero-recv rank path.
- Ran the intranode test on 8xH100 and it worked.
- Also exercised this in an external MoE model with imbalanced routing, including cases where some ranks received 0 tokens, and it worked there as well.

- [ ] Unit tests
- [x] Integration tests
- [x] Manual testing

## Checklist
- [x] I have added tests.

### DIFF
--- a/ep/bench/buffer.py
+++ b/ep/bench/buffer.py
@@ -914,7 +914,9 @@ class Buffer:
                 is_token_in_rank,
                 send_head,
             ) = handle
-            num_recv_tokens = int(rank_prefix_matrix[self.group_size - 1, self.rank].item())
+            num_recv_tokens = int(
+                rank_prefix_matrix[self.group_size - 1, self.rank].item()
+            )
             num_topk = 0
             num_scales = (
                 0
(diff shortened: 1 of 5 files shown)


YangZhou1997:
Thank you @Matt-Harvill for your contribution! I guess @MaoZiming or @zhenhuang12  would be the best one to review.

ep/bench/buffer.py
```diff
@@ -914,7 +914,7 @@ def dispatch(
                 is_token_in_rank,
                 send_head,
             ) = handle
-            num_recv_tokens = recv_src_idx.size(0)
+            num_recv_tokens = int(rank_prefix_matrix[self.group_size - 1, self.rank].item())
```
zhenhuang12: access a gpu-side tensor via `.item()` will cause  cpu-gpu sync, which may lead to performance degradation.

Matt-Harvill: Ah yes good catch

Matt-Harvill: Updated this. Decided to update both `intranode_dispatch` and `internode_dispatch` to be more explicit about this @zhenhuang12 but was only able to test intranode. I can revert the internode changes but they appear harmless

YangZhou1997:
Okay, I think GitHub Action secret cannot be passed to a forked PR, thus the CI fails. If the PR looks good to you @zhenhuang12 , we can merge and let the main run the CI.

zhenhuang12 (approved):
LGMT

PR #903, opened 2026-04-25
[EP] Fix internode dispatch deadlock and memory ordering issues

yuhuiaws:
## Summary

Fix multiple issues in the EP internode dispatch kernel (`ep/src/internode.cu`) that cause deadlocks and data corruption when running MoE training with EP>=16 over high-latency RDMA fabrics (e.g., AWS EFA):

- **Sender window bitmap spin lock livelock** (root cause): The tight `release_lock`→`acquire_lock` spin loop has no yield, causing warp starvation. Added `__nanosleep` after `release_lock`.
- **Forwarder NVL blocking wait deadlock**: The outer `while(true)` NVL slot wait blocks the Forwarder from servicing other RDMA ranks. Replaced with a non-blocking check.
- **Memory ordering for RDMA puts**: Added `__threadfence_system` before token/metadata RDMA puts and before Forwarder reads, ensuring DMA visibility.
- **Epoch tag for stale data detection**: Encode monotonic tag in upper 24 bits of `SourceMeta.is_token_in_nvl_rank_bits`; Forwarder spin-waits on tag; Receiver masks it out.
- **Forwarder round-robin fairness**: Changed early-stopping limit from `num_max_nvl_chunked_send_tokens` to `num_max_rdma_chunked_send_tokens` to prevent cross-RDMA starvation.
- **Coordinator partial head advance**: Added ~1ms stall detection for more responsive head advances.
- **Improved timeout diagnostics**: All timeout messages now include NVL state, live head readback, and sender progress.
- **Buffer config for EP=16**: Increased `rdma_recv` from 128 to 512 for both dispatch and combine to absorb EFA round-trip latency.

## Background

These issues were discovered during Qwen3-235B-A22B MoE (128 experts, top-8) training on 4×8 H200 GPUs over AWS EFA with EP=16. The internode dispatch kernel would deadlock after 2-30 iterations. Root cause analysis using diagnostic timeouts confirmed the Sender warp was stuck in the window bitmap spin lock (lines 902-930) — multiple Sender warps competing for the same per-lane lock caused livelock due to the tight spin with no yield.

## Test plan

- [x] Qwen3-235B-A22B training stable for 249+ iterations (~13 hours) on 4×8 H200 over EFA, EP=16, TP=4, PP=2, CP=2
- [x] No `SENDER_WINDOW_SPIN`, `EPOCH_TIMEOUT`, or `NVL_FULL_TIMEOUT` diagnostics fired
- [x] Throughput: 68.9 TFLOP/s/GPU (vs 66.9 TFLOP/s/GPU with NCCL alltoall baseline, ~3% faster)
- [ ] Test on InfiniBand fabric
- [ ] Test with EP=32/64

🤖 Generated with [Claude Code](https://claude.com/claude-code)


YangZhou1997:
This seems like a heavy change. What do you think cc @MaoZiming ?

MaoZiming:
@yuhuiaws Thank you! we are taking a look at this issue (internode stuck on MoE training on EFA) these days. will review

yuhuiaws:
I am not sure if the uccl-ep on EFA had tested on multiple nodes training. 
In current design, the code was easy to crash/timeout when rdma_recv was 128 for EP 32 in my megatron-lm training with uccl-ep wrapper(max token length was 20K). 
And I tried to fix the issue, more issues happened. And when I adjust rdma_recv  from 128 to 256 or 512, the issue disappeared. 
But this was just workaround not the real fix. 
And if more EP degree and more token lenght(such as 500K), I am not sure how to handle it in current design.

MaoZiming:
@yuhuiaws Yes, to run on EFA we need to tune the dispatch/combine configs (for EFA), e.g. chunk size/rcv tokens, the default setting there is on CX7. Miconfiguration can lead to deadlock. We haven't done much testing for training on EFA, we only tested training on AMD GPUs (with various NICs) cc @YangZhou1997

yuhuiaws:
@MaoZiming, Thanks for your response.
And in our GPU projects, customers will use deepep on IB/Roce with Connect-X both for training and inference. 
So I think uccl-ep on EFA may be a good workaround because deepep on EFA is not good by now. 
And I had done many tests about uccl-ep with xPxD deployment, the solution had better performance than nccl ep with xPxD.
Just for your information.

ep/src/internode.cu
```diff
@@ -702,6 +702,7 @@ __global__ void __launch_bounds__(
 
       // Issue RDMA for non-local ranks
       if (dst_rdma_rank != rdma_rank) {
+        __threadfence_system();  // ensure metadata visible to NIC DMA
```
MaoZiming: This doesn't seem to be needed (since it goes through the cpu proxy threads)

MaoZiming:
@yuhuiaws Are the `__threadfence_system` and additional waiting (e.g. https://github.com/uccl-project/uccl/pull/903/changes#diff-0bae78cf62a91d46df01ce01816fd348d46dbd04a71876788eb5986599c7981fR1239-R1266) needed for correctness? From my test, it seems changing the dispatch/combine config on EFA is enough.

yuhuiaws:
> @yuhuiaws Are the `__threadfence_system` and additional waiting (e.g. https://github.com/uccl-project/uccl/pull/903/changes#diff-0bae78cf62a91d46df01ce01816fd348d46dbd04a71876788eb5986599c7981fR1239-R1266) needed for correctness? From my test, it seems changing the dispatch/combine config on EFA is enough.

According to my tests，the related code changes were needed. And the issues may happen after many train steps, so you can try such as 100 train steps.

ep/src/internode.cu
yuhuiaws: Ok.

MaoZiming:
Summarze the main correctness issues/fixes:

### Correctness fix: stale `SourceMeta` reads in the forwarder

**Root cause.** The forwarder read per-token routing metadata
(`is_token_in_nvl_rank_bits`) from the RDMA recv ring with `ld_nc_global` — a
non-coherent, weakly-ordered load. When the sender wrapped the chunked ring
and reused a slot, the forwarder could observe **stale bytes from a prior
epoch** before the NIC's DMA for the new epoch retired into L2. `seen_bits`
would then be wrong (or zero), causing misrouted tokens, drops, or
`if (seen_bits == 0) trap()` deadlocks.

### Fix

1. **Epoch-tag the metadata.** Sender stamps the upper 24 bits of
   `is_token_in_nvl_rank_bits` with `(rdma_tail_idx + 1) & 0xFFFFFF`. Lower
   8 bits remain the NVL routing bitmap (only 8 NVL peers).
2. **Acquire-load + tag spin-wait.** Forwarder switches from `ld_nc_global`
   to `ld_acquire_sys_global`; on tag mismatch it spins with
   `ld_volatile_global` until the expected epoch is observed. Guarantees the
   bytes belong to the current epoch, not a prior reuse.
3. **Strip the tag on the final consumer.** Receive-side `SourceMeta` is
   reconstructed field-by-field and the routing field is `& 0xFF`'d.

### Deadlock fixes (what the branch name refers to)

- **Per-token NVL slot wait** in the forwarder. Previous early-stop on a full
  NVL buffer could still overshoot `head + capacity` within a chunk. Now
  blocks per-token until `tail - head < capacity`.
- **Partial RDMA-head advance** in the coordinator. Old code only advanced
  the remote head when ≥ `num_max_rdma_chunked_send_tokens` accumulated,
  deadlocking on a non-zero residual after the last full batch. Now advances
  on any positive delta after a ~1 ms stall.

### Others non-correctness-related changes

- Richer timeout `printf`s + a window-spin watchdog (diagnostics only).
- Six `__threadfence_system()` calls left commented with `Note(MaoZiming): …`
  — already covered by `atomic_set_and_commit`'s fence (proxy path) or by
  `st_release_sys_global` (NVL publish); a consumer-side fence cannot
  synchronize with NIC DMA.
- EP rank-16/32 dispatch & combine configs bump the last config field to 512
  on EFA only (gated by `_is_efa()` / `EFA_HOME`); other fabrics keep 128.

Issue #893, opened 2026-04-20
[Bug] UCCL-EP Megatron training crashes + 5× slower than NCCL-native on 4× p5en EFAv3 (DS-V3 FP8 EP=32)


AutoJunjie:
**Environmen**t:

- AWS EKS, 4× p5en.48xlarge (8× H200, 16× 400 Gb/s EFAv3 per node), 32 GPUs, us-east-2a
- `uccl==0.1.0.post6`, `ep.abi3.so` rebuilt locally with `-DUSE_DMABUF`
- Megatron-core 0.17.0 direct (torchrun, no Ray), DeepSeek-V3-like MoE
- HIDDEN=7168, 12 layers, 256 experts, topk=8, seq=4096
- EP=32 (flex dispatcher, `--moe-enable-deepep`), PP=1, TP=1, micro-bs=1, global-bs=32
- Dtype: FP8 hybrid (`--fp8-format hybrid`), low-latency mode

**Setup notes**:

Stock `uccl==0.1.0.post6` pip wheel is unusable on EFAv3 — `ibv_reg_mr_iova2` returns `ENOTSUP` because the DMABUF path is gated behind `INTEL_RDMA_NIC`. We rebuilt `ep.abi3.so` with `-DUSE_DMABUF` to get past this. To reach step 1 we also needed:

1. `ep/buffer.py:1495` — non-cached internode dispatch hardcodes `previous_event=None`, which trips `EP_HOST_ASSERT(previous_event.has_value() and async)`. Patched to `getattr(previous_event, "event", None)`.
2. A `sitecustomize.py` shim monkey-patching `deep_ep.Buffer.dispatch/.combine` to coerce `async_finish=False, allocate_on_comm_stream=False` when `previous_event is None` — Megatron 0.17's `MoEFlexTokenDispatcher` defaults to async + comm-stream, which trips the same assert without the shim.

Happy to share either patch if useful.

**Symptom**:

Training crashes intermittently (step 94–112, unpredictable), before step 120 in every run we attempted. `test_internode.py` at paper config (`--num-experts=288 --hidden=7168 --topk=8 --num-tokens=4096`, 32 ranks) always passes — dispatch NVL 57.0 GB/s (FP8) / 59.1 GB/s (BF16), combine NVL 44.1 GB/s, which meets or exceeds the paper's 54 / 43 GB/s numbers on this cluster. So the transport layer itself looks healthy; failures only appear during actual Megatron training.

[...]

Issue #878, opened 2026-04-13
[Bug] Intermittent crash during EP=16 training on EFA (p5en)


renov:
**Environment:**
- AWS EKS, 16× p5en.48xlarge (8× H200, 16× 200Gb/s EFA), 128 GPUs
- uccl-ep at commit `d27e825` (Mar 3, 2026)
- Megatron-Core SFT (verl + Ray), Qwen3-235B MoE
- EP=16 normal mode (BF16), PP=8, VPP=2

**Setup notes:**
Running under Ray (not torchrun), so we manually set `LOCAL_WORLD_SIZE=8` and `RAY_EXPERIMENTAL_NOSET_CUDA_VISIBLE_DEVICES=1` to expose all GPUs. We also patched `initialize_uccl()` to use IP-based topology via `all_gather_object` (instead of env-var `LOCAL_RANK`) and added 180s sleep per EP group after `start_dual()` — these appeared to delay crashes but did not eliminate them (e.g., 100 steps succeeded in one run, but crashed at step 15 in another with identical config).

**Symptom:**
Training crashes intermittently (step 5–100+, unpredictable). Benchmark `test_internode.py` always passes — only fails during multi-EP-group training with PP pipeline (8 EP groups of 16 GPUs, initialized sequentially).

**Master log:**

[wait_until_cmd_consumed nvl:3 cmd:4 label:-1] waiting slot=1465155
Global rank: 0, num_recv_tokens: -1, num_rdma_recv_tokens: -1
DeepEP timeout check failed: rank = 0, thread = 0, value = 0)
DeepEP timeout check failed: rank = 0, thread = 1, value = 0)
DeepEP timeout check failed: rank = 0, thread = 2, value = 0)
...

**Worker log (per-worker):**

Unhandled exception: N3c1016DistBackendErrorE. what():
[PG ID 6 PG GUID 299(PIPELINE_MODEL_PARALLEL_GROUP) Rank 5]
Process group watchdog thread terminated with exception:
Watchdog caught collective operation timeout: WorkNCCL(SeqNum=1410,
OpType=COALESCED, Timeout(ms)=1800000) ran for 1800020 milliseconds before timing out.

**Questions:**
1. Is uccl-ep expected to work with Ray-based distributed runtimes (not torchrun)? Any known caveats?
[...]

PR #904, opened 2026-04-25
[EP] Allow runtime override of CPU recv timeout via UCCL_EP_CPU_TIMEOUT_SECS

KevinZhao:
# [EP] Allow runtime override of CPU recv timeout via UCCL_EP_CPU_TIMEOUT_SECS

## Description

Allow runtime override of the CPU-side recv timeout via the new
`UCCL_EP_CPU_TIMEOUT_SECS` environment variable.

The timeout is currently a compile-time constant `NUM_CPU_TIMEOUT_SECS`
(100 seconds in non-FAST_DEBUG builds, defined in `ep/include/ep_configs.cuh:14`),
consumed at `ep/src/uccl_ep.cc` (two sites in `Buffer::low_latency_dispatch` and
`Buffer::dispatch`) to bound the `moe_recv_counter` host-side spin loop.

Long training steps (e.g. DeepSeek-V3 / Qwen3-235B Megatron runs with large
grad accumulation, heavy checkpoint I/O, or EP groups that initialize
sequentially under PP ≥ 2) can exceed the 100 s window between successive
dispatches and falsely trigger:

```
DeepEP error: CPU recv timeout
DeepEP error: timeout (dispatch CPU)
```

This was explicitly requested by @AutoJunjie in
https://github.com/uccl-project/uccl/issues/893:

> *"Is there a runtime knob for `NUM_TIMEOUT_CYCLES`, or is re-compiling the
> only path today?"*

The same symptom family also appears in
https://github.com/uccl-project/uccl/issues/878 (Qwen3-235B EP=16 intermittent
crashes on AWS EFA).

Scope note: this PR only touches the CPU-side spin timeout. The GPU-side
`NUM_TIMEOUT_CYCLES` is untouched.

Fixes first half of #893 (runtime knob for CPU timeout).
Related to #878 (similar symptom on EFA).

## Type of Change
- [x] Bug fix

## Design

- New helper `get_cpu_timeout_secs(int fallback_secs)` in `ep/include/common.hpp`,
  placed alongside the four existing env helpers
  (`get_max_inflight_bytes`, `get_max_inflight_low_latency`,
  `get_max_inflight_normal`, `get_aggressive_atomic_enabled`) to match the
  established pattern.
- `fallback_secs` is passed as an argument so `common.hpp` does not need to
  include `ep_configs.cuh`; callers pass the compile-time constant
  `NUM_CPU_TIMEOUT_SECS` explicitly.  The static-local cache remains valid
  because callers always pass the same compile-time value.
- Parses with `strtol` + validation (not `atoi`):
  - env unset / empty        → silent fallback to `NUM_CPU_TIMEOUT_SECS`
  - valid positive int       → use it, capped at `INT_MAX`
  - 0 / negative / garbage   → fallback + one-time `[UCCL] Warning: ...` to stderr
- Parsed once per process via C++11 thread-safe static local (magic static),
  matching the existing env-helper pattern.

## How Has This Been Tested?

- [x] **Format**: `./format.sh` (clang-format-14, black) produces no diff on the
      two changed files.
- [x] **Static check**: `clang-format-14 --dry-run --Werror ep/include/common.hpp
      ep/src/uccl_ep.cc` clean.
- [ ] **Build**: `bash build.sh cu12 ep --install` on an H200 host — I do not
      currently have a p5en node warm; planning to run this together with a
      full smoke pass on Stage 5, and will attach the log as a follow-up
      comment before merge.
- [ ] **Unit tests**: none added.  The env is parsed lazily via a
      static-local cache, so same-process pytest parametrization does not
      work — a proper test needs `subprocess` isolation across cases.
      Happy to add `ep/tests/test_cpu_timeout_env.py` in a follow-up commit
      if desired; otherwise the existing microbench (`test_low_latency.py`)
      will exercise the unset-env path.
- [ ] **Manual stress**: Megatron run with `UCCL_EP_CPU_TIMEOUT_SECS=600` to
      confirm no false timeout during heavy grad-accum steps — planned as
      part of Stage 5 validation; will attach results.

## Checklist
- [ ] I have run `build.sh cu12 ep --install` to verify compilation.  *(see
      "How Has This Been Tested?" above — GPU host pending)*
- [ ] I have updated the documentation.  *(README env-var table update will
      land as a separate commit once the API is settled in review.)*
- [ ] I have added tests.  *(see notes above.)*

## Notes for reviewers

- Env name choice: went with `UCCL_EP_CPU_TIMEOUT_SECS` (EP-specific) to
  mirror the existing `UCCL_EP_*` naming for EP-module knobs
  (e.g. `UCCL_EP_ENABLE_AGGRESSIVE_ATOMIC`).  Happy to rename if you'd
  prefer `UCCL_CPU_TIMEOUT_SECS` (repo-wide) or `DEEPEP_CPU_TIMEOUT_SECS`
  (align with the error string).

- Validation policy: silent fallback for unset/empty vs stderr warning for
  garbage/negative/zero.  The rationale is that unset is the normal path
  for 99% of users, while garbage/negative is almost always a user error
  worth surfacing.  Can be toned down to fully silent or switched to
  `glog` if preferred.

/cc @MaoZiming @YangZhou1997

### DIFF
--- a/ep/include/common.hpp
+++ b/ep/include/common.hpp
@@ -3,6 +3,8 @@
 
 #include "util/gpu_rt.h"
 #include <atomic>
+#include <cerrno>
+#include <climits>
 #include <cstdint>
 #include <cstdio>
 #include <cstdlib>
@@ -146,6 +148,43 @@ static inline bool get_aggressive_atomic_enabled() {
     return env ? static_cast<uint32_t>(atoi(env)) : 0;
   }();
   return val != 0;
+}
+
+// Runtime override for the CPU-side recv timeout.
+// Long training steps (e.g. Megatron-LM with large grad accumulation, heavy
+// checkpoint I/O, or EP groups initializing sequentially under PP >= 2) can
+// exceed the compile-time `NUM_CPU_TIMEOUT_SECS` default (100 s) and falsely
+// trigger "DeepEP error: CPU recv timeout" / "timeout (dispatch CPU)".
+// Set UCCL_EP_CPU_TIMEOUT_SECS to a larger integer (seconds) to widen the
+// window.
+//
+// `fallback_secs` should be the compile-time default (NUM_CPU_TIMEOUT_SECS)
+// from ep_configs.cuh; passing it as an argument lets us keep common.hpp free
+// of that include.  Only the first call's fallback takes effect because of the
+// static-local cache; callers must therefore pass a consistent value.
+//
+// Parsing rules (match the style of get_max_inflight_* above):
+//   - env unset or empty         -> fallback
+//   - env = valid positive int   -> use it (capped at INT_MAX)
+//   - env = 0, negative, garbage -> fallback, warn once to stderr
+static inline int get_cpu_timeout_secs(int fallback_secs) {
+  static int const val = [fallback_secs]() -> int {
+    char const* env = getenv("UCCL_EP_CPU_TIMEOUT_SECS");
+    if (env == nullptr || *env == '\0') return fallback_secs;
+    char* end = nullptr;
+    errno = 0;
+    long v = strtol(env, &end, 10);
+    if (errno != 0 || end == env || *end != '\0' || v <= 0) {
+      fprintf(stderr,
+              "[UCCL] Warning: invalid UCCL_EP_CPU_TIMEOUT_SECS='%s', "
+              "using default %d seconds\n",
+              env, fallback_secs);
+      return fallback_secs;
+    }
+    if (v > INT_MAX) v = INT_MAX;
+    return static_cast<int>(v);
+  }();
+  return val;
 }
 
 #endif  // COMMON_HPP
--- a/ep/include/common.hpp
+++ b/ep/include/common.hpp
@@ -155,18 +155,6 @@ static inline bool get_aggressive_atomic_enabled() {
 // checkpoint I/O, or EP groups initializing sequentially under PP >= 2) can
 // exceed the compile-time `NUM_CPU_TIMEOUT_SECS` default (100 s) and falsely
 // trigger "DeepEP error: CPU recv timeout" / "timeout (dispatch CPU)".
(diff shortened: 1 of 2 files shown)


KevinZhao:
Follow-up commit `a7eb743e` addresses two self-review items before external review lands:

1. **Consistency**: moved `get_cpu_timeout_secs` from file scope in `uccl_ep.cc` into `ep/include/common.hpp` alongside the other `get_max_inflight_*` / `get_aggressive_atomic_enabled` helpers.  The fallback value is passed as an argument so `common.hpp` doesn't need to pull in `ep_configs.cuh`.

2. **Robust parsing**: swapped `atoi()` for `strtol()` with explicit validation. Invalid input (`"abc"`, `"-5"`, `"0"`) now emits a one-time `[UCCL] Warning: invalid UCCL_EP_CPU_TIMEOUT_SECS='...', using default N seconds` to stderr instead of silently falling back, so users can see they mistyped. Also capped at `INT_MAX` to avoid overflow on absurdly large values.

Both changes are non-functional relative to the fast path. PR body updated accordingly; README env-var table update will come as a separate commit once the final API naming is settled.

ep/include/common.hpp
```diff
@@ -148,4 +150,41 @@ static inline bool get_aggressive_atomic_enabled() {
...
+//
+// `fallback_secs` should be the compile-time default (NUM_CPU_TIMEOUT_SECS)
+// from ep_configs.cuh; passing it as an argument lets us keep common.hpp free
+// of that include.  Only the first call's fallback takes effect because of the
+// static-local cache; callers must therefore pass a consistent value.
+//
+// Parsing rules (match the style of get_max_inflight_* above):
+//   - env unset or empty         -> fallback
+//   - env = valid positive int   -> use it (capped at INT_MAX)
+//   - env = 0, negative, garbage -> fallback, warn once to stderr
```
YangZhou1997: Can we remove these lines of comments, which read unnecessarily complex

YangZhou1997 (commented):
Thank you for the contribution

YangZhou1997 (approved):

PR #936, opened 2026-05-12
[P2P] fixing segv when using rdma inline

YangZhou1997:
## Description

Fixes # (issue)

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [x] Unit tests
- [ ] Integration tests
- [ ] Manual testing

## Checklist
- [ ] I have added tests.

### DIFF
--- a/p2p/engine.h
+++ b/p2p/engine.h
@@ -247,7 +247,7 @@ class Endpoint {
   static constexpr int kMaxNumGPUs = 8;
   static constexpr size_t kIpcAlignment = 1ul << 20;
   static constexpr size_t kIpcSizePerEngine = 1ul << 20;
-  static constexpr int kMaxInflightOps = 8;  // Max 8 concurrent Ops
+  static constexpr int kMaxInflightOps = 128;  // Max 128 concurrent Ops
   static constexpr size_t ShmRingDefaultElemCnt = 16;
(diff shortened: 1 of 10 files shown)


YangZhou1997:
@monopodium can you give a review? this removes RDMA inline send as inline data cannot be delivered to GPU memory, causing SEGV

p2p/engine.h
```diff
@@ -247,7 +247,7 @@ class Endpoint {
   static constexpr int kMaxNumGPUs = 8;
   static constexpr size_t kIpcAlignment = 1ul << 20;
   static constexpr size_t kIpcSizePerEngine = 1ul << 20;
-  static constexpr int kMaxInflightOps = 8;  // Max 8 concurrent Ops
+  static constexpr int kMaxInflightOps = 128;  // Max 128 concurrent Ops
```
monopodium: Does this change because of performance considerations? Have you checked if this will exceed the NIC capability like CQ/SQ?

monopodium (approved):

p2p/engine.h
YangZhou1997: Yes, it is mostly for performance tuning. It should not hit the CQ/SQ limit, but let me remove it in this PR and do it in another one

PR #940, opened 2026-05-13
Fix deadlock in internode collective functions when num_tokens == 0

### LINKED ISSUE #935: Assertion Failures When Used With SGLang v0.5.11 DeepEP Integration
# UCCL-EP: Assertion Failures When Used With SGLang v0.5.11 DeepEP Integration

## Summary

UCCL-EP (commit `605e8f2b`) crashes with null pointer assertions when integrated with SGLang v0.5.11's DeepEP backend. Both `low_latency` and `normal` dispatch modes fail with different assertion errors during inference. The server starts successfully (CUDA graph capture, DeepGEMM JIT, warmup all pass for `normal` mode) but crashes on the first real inference request.

## Severity

High — completely blocks SGLang + UCCL-EP usage on multi-node EP deployments.

## Environment

| Component | Version |
|-----------|---------|
| SGLang | 0.5.11 |
| UCCL-EP | commit `605e8f2b3f0d4f73b9f7291130d8574840a241fe` |
| PyTorch | 2.11.0+cu130 |
| CUDA | 13.0 |
| NVIDIA Driver | 580.x |
| NCCL | 2.30.4-1 |
| EFA Installer | 1.48.0 |

## Hardware

- Instance: 2× AWS p5en.48xlarge
- GPU: 8× NVIDIA H200 (141 GiB) per node (16 total)
- Network: 16× EFA devices per node (Libfabric/GDRDMA confirmed working)

### Configuration

```bash
# DeepEP config (deepep_config.json)
{
  "normal_dispatch": {
    "num_sms": 24,
    "num_max_nvl_chunked_send_tokens": 16,
    "num_max_nvl_chunked_recv_tokens": 512,
    "num_max_rdma_chunked_send_tokens": 16,
    "num_max_rdma_chunked_recv_tokens": 512
  },
  "normal_combine": {
    "num_sms": 24,
    "num_max_nvl_chunked_send_tokens": 16,
    "num_max_nvl_chunked_recv_tokens": 512,
    "num_max_rdma_chunked_send_tokens": 16,
    "num_max_rdma_chunked_recv_tokens": 512
  }
}
```

### Launch command (head node, node-rank=0)

```bash
export NVSHMEM_REMOTE_TRANSPORT=libfabric
export NVSHMEM_LIBFABRIC_PROVIDER=efa
export FI_PROVIDER=efa
export FI_EFA_USE_DEVICE_RDMA=1
export SGLANG_ENABLE_JIT_DEEPGEMM=1
export SGLANG_DEEPEP_NUM_MAX_DISPATCH_TOKENS_PER_RANK=512

python3 -m sglang.launch_server \
  --model-path deepseek-ai/DeepSeek-V3-0324 \
  --tp-size 16 \
  --ep-size 16 \
  --dp-size 16 \
  --nnodes 2 \
  --node-rank 0 \
  --dist-init-addr "${MY_IP}:5000" \
  --trust-remote-code \
  --mem-fraction-static 0.85 \
  --moe-dense-tp-size 1 \
  --chunked-prefill-size 32768 \
  --cuda-graph-bs 128 \
  --page-size 256 \
  --attention-backend fa3 \
  --ep-num-redundant-experts 0 \
  --ep-dispatch-algorithm dynamic \
  --enable-dp-attention \
  --enable-dp-lm-head \
  --moe-a2a-backend deepep \
  --deepep-mode normal \
  --deepep-config /tmp/deepep_config.json \
  --host 0.0.0.0 \
  --port 30000 \
  --watchdog-timeout 1000000
```

### Trigger

Send any completion request after the server reports "fired up and ready to roll":

```bash
curl http://localhost:30000/v1/completions \
  -H "Content-Type: application/json" \
  -d '{"model": "deepseek-ai/DeepSeek-V3-0324", "prompt": "The capital of France is", "max_tokens": 20, "temperature": 0}'
```

### Error 1: `normal` mode

The server starts, loads model, captures CUDA graphs, completes warmup, and reports ready. On the first real inference request, it crashes:

```
RuntimeError: Failed: Assertion error /opt/uccl/ep/src/uccl_ep.cc:521 'topk_idx_ptr != 0'
```

Full traceback shows the error in `get_dispatch_layout`:
```python
    event = self.runtime.get_dispatch_layout(
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Failed: Assertion error /opt/uccl/ep/src/uccl_ep.cc:521 'topk_idx_ptr != 0'
```

This then cascades into a `DeepEP error: timeout (dispatch CPU)` on other ranks.

### Error 2: `low_latency` mode

With `--deepep-mode low_latency`:

**First attempt** (cuda-graph-bs=256, SGLANG_DEEPEP_NUM_MAX_DISPATCH_TOKENS_PER_RANK=128):
```
Exception: Capture cuda graph failed: Failed: Assertion error /opt/uccl/ep/src/uccl_ep.cc:1202 'x_rows <= num_max_dispatch_tokens_per_rank'
```
This was fixed by increasing `SGLANG_DEEPEP_NUM_MAX_DISPATCH_TOKENS_PER_RANK=512` and reducing `cuda-graph-bs=128`.

**Second attempt** (cuda-graph-bs=128, SGLANG_DEEPEP_NUM_MAX_DISPATCH_TOKENS_PER_RANK=512):
```
RuntimeError: Failed: Assertion error /opt/uccl/ep/src/uccl_ep.cc:1197 'x_ptr != 0 && topk_idx_ptr != 0'
```
Crash on first decode step after warmup prefill.

## Expected Behavior

UCCL-EP should handle the tensors passed by SGLang's DeepEP integration without assertion failures. If SGLang passes empty/null `topk_idx` tensors for ranks that receive zero routed tokens (an edge case in sparse MoE routing), UCCL-EP should handle this gracefully (skip dispatch for that rank) rather than asserting.

## Analysis

1. The warmup prefill (256 tokens) works. This suggests the issue is specific to the decode path (1 token per request) where token routing may leave some ranks with zero dispatched tokens.

2. SGLang's MoE layer appears to pass `topk_idx=None` or an empty tensor when a rank has no tokens routed to its experts. The UCCL-EP assertion at line 521 (`uccl_ep.cc`) rejects this.

3. Note: UCCL-EP commit `09aada2` (Mar 28) is described as "[EP] Remove redundant check to allow for empty topk_idx tensors" — this is already included in our 605e8f2 build. However, it appears the fix was incomplete or SGLang triggers a different code path.

4. Newer commits on the main branch (May 9, 2026) may address this:
   - `9a239f0`: "[EP] Accept DeepEP-compatible overlap kwargs in low_latency_combine (stub)"
   - `55da9f8`: "[EP] Allow high-throughput + low-latency Buffers to coexist in one process"

## Workaround

None currently. UCCL-EP + SGLang DeepEP is non-functional. Only the NCCL backend (`--moe-a2a-backend none`) works with SGLang EP.

## Additional Context

- vLLM v0.20.1 with the same UCCL-EP commit (605e8f2) works correctly with `deepep_low_latency` on the same model and hardware. This suggests SGLang's DeepEP integration calls the UCCL-EP API differently from vLLM's.

- The UCCL benchmark scripts (ep/bench/sglang/) reference launch patterns that may assume a newer UCCL-EP version or specific SGLang patches.

- SGLang's NCCL EP path (without DeepEP) works correctly but is 2.3× slower than vLLM's allgather_reducescatter baseline (107ms vs 46.7ms TPOT at EP=16 on DeepSeek-V3-0324).

## Requested Fix

1. Handle null/empty `topk_idx` tensors gracefully in both `get_dispatch_layout` (line 521) and the low-latency dispatch path (line 1197).
2. Confirm which UCCL-EP commit is required for SGLang v0.5.11 compatibility.
3. Document the minimum `SGLANG_DEEPEP_NUM_MAX_DISPATCH_TOKENS_PER_RANK` required for various `cuda-graph-bs` settings.

## Related

- UCCL issue #923 (silent deadlock with EP=16 — fixed in 605e8f2)
- SGLang UCCL bench scripts: `ep/bench/sglang/deepseek_r1_uep_ll.sh`

pbelevich:
### Problem

When SGLang uses DP-attention with DeepEP `normal` mode, some EP ranks may have 0 tokens to dispatch in a given step. The existing early-return paths in `internode_prepare`, `internode_dispatch`, and `internode_combine` would skip the collective `cached_notify` / `notify_dispatch` calls, causing remote ranks to spin forever on their CPU wait loops polling `moe_recv_counter`. This manifests as:

```
RuntimeError: DeepEP error: timeout (dispatch CPU)
rank=0, thread=0, value=0
```

while threads 1–7 report `value=1024`, indicating they received notifications from some ranks but not from the zero-token rank that early-returned.

### Root Cause

`internode_prepare`, `internode_dispatch`, and `internode_combine` all contain collective signaling (via `notify_dispatch`, `cached_notify`, and the dispatch/combine CUDA kernels). When a rank skips these calls, remote ranks deadlock waiting for notifications that never arrive.

### Fix

1. **`internode_prepare`**: Remove the early-return. Relax assertion from `num_tokens > 0` to `num_tokens >= 0`. The rank always participates in `notify_dispatch` even with 0 tokens.

2. **`internode_dispatch`**: Remove the early-return. Relax assertion to `num_tokens >= 0`. Guard data-pointer assertions behind `if (num_tokens > 0)`. Use null-safe ternary casts for `x_ptr`, `recv_x_ptr`, and `is_token_in_rank_ptr` so the dispatch kernel receives `nullptr` (safe because it iterates 0 times over the token dimension).

3. **`internode_combine`**: Remove the early-return. Guard data-pointer assertions behind `if (num_combined_tokens > 0)`. Use null-safe ternary casts for `combined_x_ptr`, `is_combined_token_in_rank_ptr`, `x_ptr`, and `src_meta_ptr`.

Non-collective functions (`get_dispatch_layout`, `low_latency_dispatch`, `low_latency_combine`) retain their early-returns since they use RDMA point-to-point operations that don't require collective participation.

### Testing

Verified on 2-node (16×H200) deployment with SGLang v0.5.11 serving DeepSeek-V3-0324, EP=16, DP-attention, DeepEP `normal` mode:

- Server starts and completes CUDA graph capture (previously succeeded)
- **Single request completes successfully** (previously deadlocked here)
- 3 concurrent requests complete successfully
- 32 requests at 4 RPS complete with 0 failures

### Performance

| Workload | RPS | Median TPOT | vs NCCL baseline |
|----------|-----|-------------|-----------------|
| 512×128 | 1 | 17.2ms | 6.2× faster |
| 1024×256 | 2 | 31.5ms | 3.4× faster |
| 1024×256 | 4 | 34.7ms | 3.1× faster |

NCCL EP baseline on same hardware/model: 107ms median TPOT.

### Environment

- SGLang v0.5.11
- PyTorch 2.11.0+cu130
- CUDA 13.0
- NCCL 2.30.4
- 2× p5en.48xlarge (16× H200, EFA)
- Model: deepseek-ai/DeepSeek-V3-0324 (FP8)

Fixes #935

### DIFF
--- a/ep/src/uccl_ep.cc
+++ b/ep/src/uccl_ep.cc
@@ -908,26 +908,6 @@ class Buffer {
                     bool allocate_on_comm_stream,
                     std::uintptr_t compute_stream_ptr) {
     nb::gil_scoped_release release;
-    // Handle empty token case: when num_tokens == 0, this rank has nothing
-    // to dispatch. Return zeros without launching any communication.
-    if (num_tokens == 0) {
-      auto compute_stream = reinterpret_cast<cudaStream_t>(compute_stream_ptr);
-      static_cast<void>(allocate_on_comm_stream);
-      if (previous_event.has_value()) {
-        stream_wait(comm_stream, previous_event.value());
-      } else {
-        stream_wait(comm_stream, compute_stream);
-      }
-      int const num_local_experts = num_experts / num_ranks;
-      std::vector<int> empty_expert_counts(num_local_experts, 0);
-      std::optional<EventHandle> event;
-      if (async) {
-        event = EventHandle(comm_stream);
-      } else {
-        stream_wait(compute_stream, comm_stream);
-      }
-      return {0, 0, empty_expert_counts, event};
-    }
     EP_HOST_ASSERT(num_tokens_per_rank_ptr != 0);
     EP_HOST_ASSERT(num_tokens_per_rdma_rank_ptr != 0);
     EP_HOST_ASSERT(num_tokens_per_expert_ptr != 0);
@@ -936,7 +916,9 @@ class Buffer {
     EP_HOST_ASSERT(recv_rdma_rank_prefix_sum_ptr != 0);
     EP_HOST_ASSERT(gbl_channel_prefix_matrix_ptr != 0);
     EP_HOST_ASSERT(recv_gbl_rank_prefix_sum_ptr != 0);
-    EP_HOST_ASSERT(hidden > 0 && num_experts > 0);
+    // Allow num_tokens == 0: DP-attention ranks may have no tokens to
+    // dispatch, but must still participate in the collective notification.
+    EP_HOST_ASSERT(num_tokens >= 0 && hidden > 0 && num_experts > 0);
     EP_HOST_ASSERT(config.num_sms % 2 == 0);
     EP_HOST_ASSERT(0 < get_num_rdma_ranks() &&
                    get_num_rdma_ranks() <= NUM_MAX_RDMA_PEERS);
@@ -1034,29 +1016,18 @@ class Buffer {
       std::uintptr_t send_rdma_head_ptr, std::uintptr_t send_nvl_head_ptr,
       std::optional<EventHandle>& previous_event, bool async,
       bool allocate_on_comm_stream, std::uintptr_t compute_stream_ptr) {
-    // Handle empty token case: nothing to dispatch from this rank.
-    if (num_tokens == 0) {
-      auto compute_stream = reinterpret_cast<cudaStream_t>(compute_stream_ptr);
-      static_cast<void>(allocate_on_comm_stream);
-      if (previous_event.has_value()) {
-        stream_wait(comm_stream, previous_event.value());
-      } else {
-        stream_wait(comm_stream, compute_stream);
-      }
-      std::optional<EventHandle> event;
-      if (async) {
-        event = EventHandle(comm_stream);
-      } else {
-        stream_wait(compute_stream, comm_stream);
-      }
-      return event;
+    // Allow num_tokens == 0: DP-attention ranks may have no tokens to
+    // dispatch, but must still participate in the collective cached_notify
+    // and dispatch so that remote ranks are not left spinning.
+    EP_HOST_ASSERT(num_tokens >= 0 && hidden > 0);
+    if (num_tokens > 0) {
+      EP_HOST_ASSERT(x_ptr != 0 && recv_x_ptr != 0 &&
+                     is_token_in_rank_ptr != 0);
     }
-    EP_HOST_ASSERT(x_ptr != 0 && recv_x_ptr != 0 && is_token_in_rank_ptr != 0);
     EP_HOST_ASSERT(rdma_channel_prefix_matrix_ptr != 0);
     EP_HOST_ASSERT(recv_rdma_rank_prefix_sum_ptr != 0);
     EP_HOST_ASSERT(gbl_channel_prefix_matrix_ptr != 0);
     EP_HOST_ASSERT(recv_gbl_rank_prefix_sum_ptr != 0);
-    EP_HOST_ASSERT(num_tokens > 0 && hidden > 0);
     EP_HOST_ASSERT(config.num_sms % 2 == 0);
 
     int const num_channels = config.num_sms / 2;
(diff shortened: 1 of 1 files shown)


YangZhou1997:
Great @pbelevich! I am also looking at the issue, but great to see your PR. Can you run ./format.sh btw?

YangZhou1997 (approved):
LGTM

PR #957, opened 2026-05-22
[EP] Fix dual-batch overlap issues

MaoZiming:
## Summary
End-to-end fix for uccl-EP's DBO / two-batch-overlap (TBO) path under sglang DP-attention. Before this PR, launching sglang with `--enable-two-batch-overlap` and `--enable-dp-attention` against a uccl-EP build would silently kill a DP worker on the very first inference request, or sometimes get stuck.

### Patches to UCCL-EP
- **`ep/src/uccl_ep.cc`** — relax the `is_token_in_rank_ptr != 0` assertion in `internode_prepare` when `num_tokens == 0`. Zero-element PyTorch tensors have `data_ptr() == 0`, so DP-attention ranks with no tokens to route legitimately pass a null pointer here.
- **`ep/src/internode.cu`** — add an explicit empty-rank fast path in `notify_dispatch`. When `num_tokens == 0` we unconditionally zero-write the `rdma_channel_prefix_matrix` / `gbl_channel_prefix_matrix` slots for `dst_rdma_rank` and return, so the kernel never dereferences `is_token_in_rank` on the empty-rank path.
- **`ep/bench/buffer.py`** — mirror DeepEP's C++ `record_stream` discipline in Python:
  - New `_record_stream_safe(tensors, stream)` helper that tags every CUDA tensor in a (possibly nested) iterable, skipping `None`.
  - Move the `alloc_ctx = torch.cuda.stream(get_comm_stream())` block earlier so the `rdma_channel_prefix_matrix` / `gbl_channel_prefix_matrix` / `recv_*_prefix_sum` tensors are allocated *on comm_stream*, which makes the caching allocator tie their storage lifetime to comm_stream rather than to the default compute stream.
  - Pad `send_rdma_head` to `max(x.size(0), 1)` rows so its `data_ptr()` is never null on the DP-attention empty-micro-batch case.
### Sglang patch 
- This is mostly to handle incorrect splitting of micro-batches, or when there is two few sequences.    DeepEP likely will hit the exact same bug, because the crash happens before any MoE dispatcher runs.
- Tested against sglang **v0.5.9** (SHA **bbe9c7eeb520b0a67e92d133dfc137a3688dc7f2** ). Apply from the root of any sglang checkout with `git apply ep/bench/sglang/fix-tbo-empty-microbatch.patch`.

## Validation
Two-node g5/p5 cluster, EFA + DeepEP normal mode:
```bash
# Apply the sglang patch into the vendored sglang tree
cd $SGLANG_ROOT && git apply $UCCL/ep/bench/sglang/fix-tbo-empty-microbatch.patch
# Build & install uccl-EP
cd $UCCL/ep && make -j install
# Node 0 / Node 1
bash ep/bench/sglang/run_dbo_sglang.sh 0
bash ep/bench/sglang/run_dbo_sglang.sh 1
```

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [x] Unit tests
- [ ] Integration tests
- [x] Manual testing

## Checklist
- [ ] I have added tests.


YangZhou1997 (approved):
@MaoZiming can you maybe document the specific SHA hash of the sglang for this patch?

MaoZiming:
@YangZhou1997 Sounds good. Added in the description.

PR #984, opened 2026-06-03
[EP] Close residual zero-token assertion gaps in intranode + LL paths

xyxiang7:
## Summary

After #898 (intranode zero-token wrapper + `intranode_dispatch` `num_recv_tokens` assert) and #940 (internode + LL early-returns), three crash sites and one hook-handling gap remain for DP-attention ranks with no tokens to route. This PR closes them with minimal additions.

### What still crashes today

| Path | Site | Why upstream doesn't cover it |
|---|---|---|
| `Buffer.dispatch` non-cached → `intranode_prepare` | `num_tokens > 0`, `is_token_in_rank_ptr != 0` | #898 didn't touch `intranode_prepare` |
| `Buffer.dispatch` → `intranode_dispatch` | `x_ptr`, `is_token_in_rank_ptr`, `send_head_ptr` != 0 and `num_tokens > 0` | #898 only relaxed the `num_recv_tokens` assert; send-side stays strict |
| `Buffer.combine` → `intranode_combine` | `send_head_ptr != 0` | #898 dummies `x`/`src_idx`/`topk_weights` to ≥1 row in Python, but not `send_head` (sized by `num_recv_tokens`) |
| `Buffer.low_latency_{dispatch,combine}` | `hook is None` | #940 returns `hook=None` from the runtime's zero-token early-return; SGLang's caller does `hook() if return_recv_hook` and crashes on `None` |

### Changes

**`ep/src/uccl_ep.cc`** — gate three assertion blocks on the relevant size dimension rather than relax them unconditionally, so accidental nulls still trip in the non-zero-token path:

- `intranode_prepare`: `num_tokens > 0` → `>= 0`; `is_token_in_rank_ptr != 0` only if `num_tokens > 0`.
- `intranode_dispatch`: `num_tokens > 0` → `>= 0`; `x_ptr`/`is_token_in_rank_ptr`/`send_head_ptr` only if `num_tokens > 0`.
- `intranode_combine`: `send_head_ptr != 0` only if `num_recv_tokens > 0`. Other pointers stay strict — they're already kept non-zero by #898's Python-side dummies.

**`ep/bench/buffer.py`** — substitute a no-op `hook` in `low_latency_dispatch` / `low_latency_combine` when the runtime returns `hook=None` and the caller asked for `return_recv_hook=True`.

```
ep/bench/buffer.py |  9 +++++++++
ep/src/uccl_ep.cc  | 34 ++++++++++++++++++++++++++++------
2 files changed, 37 insertions(+), 6 deletions(-)
```

### Why minimal asserts vs. wholesale relaxation

For the C++ side I gate on the size dimension that controls each pointer's allocation, mirroring how #940 handled the internode and LL paths. Pointers stay required when the size says they should be present, so a real `nullptr` bug still trips the assert in the common path.

### DIFF
--- a/ep/bench/buffer.py
+++ b/ep/bench/buffer.py
@@ -438,10 +438,8 @@ class Buffer:
             packed_recv_x_scales_storage,
             cumulative_local_expert_recv_stats,
         )
-        # When the runtime takes the zero-token early-return path it returns
-        # hook=None. SGLang's caller does `hook() if return_recv_hook` and
-        # crashes on None, so substitute a no-op so idle DP-attention ranks
-        # participate in the collective without raising.
+        # The zero-token early-return path returns hook=None; substitute a
+        # no-op so callers can unconditionally invoke hook().
         if return_recv_hook and hook is None:
             hook = lambda: None
         return (
--- a/ep/bench/buffer.py
+++ b/ep/bench/buffer.py
@@ -438,6 +438,12 @@ class Buffer:
             packed_recv_x_scales_storage,
             cumulative_local_expert_recv_stats,
         )
+        # When the runtime takes the zero-token early-return path it returns
+        # hook=None. SGLang's caller does `hook() if return_recv_hook` and
+        # crashes on None, so substitute a no-op so idle DP-attention ranks
+        # participate in the collective without raising.
+        if return_recv_hook and hook is None:
+            hook = lambda: None
         return (
             (packed_recv_x, packed_recv_x_scales) if use_fp8 else packed_recv_x,
             packed_recv_count,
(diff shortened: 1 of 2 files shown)


ep/bench/buffer.py
```diff
@@ -438,6 +438,12 @@ def low_latency_dispatch(
             packed_recv_x_scales_storage,
             cumulative_local_expert_recv_stats,
         )
+        # When the runtime takes the zero-token early-return path it returns
+        # hook=None. SGLang's caller does `hook() if return_recv_hook` and
```
MaoZiming: I think this comment is a bit too use-case specific. Maybe just mention "zero-token early-return path".

MaoZiming (approved):
Thank you! LGTM

ep/bench/buffer.py
xyxiang7: Fixed. Thanks!

PR #985, opened 2026-06-03
[EP] Support distinct HT barrier local ranks

eicherseiji:
## Summary

This draft separates the CUDA/NIC-facing local rank from the node-local slot used by the high-throughput local barrier.

In launchers such as vLLM Ray data-parallel serving, each worker actor can see exactly one GPU. In that topology `torch.cuda.current_device()` is `0` for every actor on a node, while UCCL high-throughput still needs unique local barrier slots `0..N-1`. Reusing the CUDA-visible local rank for the HT barrier causes all local actors to write slot `0`, leaving the node leader waiting for the remaining slots.

The patch adds an optional `barrier_local_rank` to the Python-facing `uccl.ep.Proxy` constructor and threads it to `Proxy::Config`. The existing `local_rank` remains the CUDA/NIC/affinity rank; only the HT local-barrier context uses `barrier_local_rank` when provided.

## Validation

Validated on an Anyscale workspace with two 8x H100 nodes using SkyRL/vLLM 0.20.2 and DeepSeek-V3-0324 dummy weights.

- Built UCCL EP against Torch 2.11.0+cu128.
- Verified imports on both GPU workers.
- Low-latency control: `deepep_low_latency`, DP=16, local DP=8, TP=1 reached readiness and returned a dummy completion.
- High-throughput test: `deepep_high_throughput`, DP=16, local DP=8, TP=1 reached readiness and returned a dummy completion.
- Prior to this split, the same HT configuration failed with `RuntimeError: DeepEP error: timeout (dispatch CPU)`.

Example debug line from the Ray/vLLM topology:

```text
[UCCL_RANK_SPLIT] rank=11 local_rank=0 barrier_local_rank=3 node_idx=1 num_nodes=2 is_intranode=False
```

## Notes

This is intentionally opened as a draft. The API shape is small and backward-compatible, but I expect maintainers may want a different naming/API boundary for the barrier slot before this is ready to merge.

### DIFF
--- a/ep/bench/utils.py
+++ b/ep/bench/utils.py
@@ -573,15 +573,6 @@ def initialize_uccl(
         num_nodes,
         detected_is_intranode,
     ) = detect_group_topology(group)
-    if os.environ.get("UCCL_DEBUG_RANK_SPLIT") == "1":
-        print(
-            "[UCCL_RANK_SPLIT] "
-            f"rank={rank} local_rank={local_rank} "
-            f"barrier_local_rank={barrier_local_rank} "
-            f"node_idx={node_idx} num_nodes={num_nodes} "
-            f"is_intranode={detected_is_intranode}",
-            flush=True,
-        )
     if is_intranode is None:
         is_intranode = detected_is_intranode
     elif is_intranode and not detected_is_intranode:
--- a/ep/bench/utils.py
+++ b/ep/bench/utils.py
@@ -99,12 +99,13 @@ def _gather_peer_ips(group):
     return ips
 
 
-def detect_group_topology(group: dist.ProcessGroup) -> Tuple[int, int, int, bool]:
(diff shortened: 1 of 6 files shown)


ep/bench/utils.py
```diff
@@ -567,9 +566,22 @@ def initialize_uccl(
...
-        group
-    )
+    (
+        local_rank,
+        barrier_local_rank,
+        node_idx,
+        num_nodes,
+        detected_is_intranode,
+    ) = detect_group_topology(group)
+    if os.environ.get("UCCL_DEBUG_RANK_SPLIT") == "1":
```
MaoZiming: Can you remove the debug lines? Thanks!

MaoZiming (approved):
Thank you! LGTM

YangZhou1997:
@eicherseiji can you run `./format.sh` to fix format?

eicherseiji:
Updated per review feedback, thanks!

PR #989, opened 2026-06-08
[EP] Split proxy CUDA device index from NIC local rank

eicherseiji:
Builds on #985.

The EP proxy used a single `local_rank` for three separate concepts:

- CUDA device selection
- NIC/NUMA affinity and CPU pinning
- HT local barrier indexing

Those can diverge under the vLLM Ray v2 executor path. In the SkyRL/GLM-4.7 recipe, workers keep full-node `CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7`, while Ray/vLLM rank metadata can still report `local_rank=0`. Using one rank value for all three purposes can therefore select the right CUDA-visible device in one path while using the wrong physical GPU rank for NIC/NUMA affinity or barrier indexing in another.

This PR splits the proxy inputs into:

- `device_index`: index into the process-local `CUDA_VISIBLE_DEVICES` view, used for `cudaSetDevice` and atomics checks
- `nic_local_rank`: physical GPU rank, derived from the CUDA device PCI BDF, used for NIC/NUMA affinity and CPU pinning
- `barrier_local_rank`: dense local HT barrier slot, ordered by physical GPU BDF

All three default to `local_rank`, so existing callers keep their current behavior. The Python topology helper also validates that two ranks on the same node do not map to the same physical GPU rank.

## Validation

Tested with vLLM Ray v2 executor backend + GLM-4.7 dummy load across two 8xH100 nodes with EFA:

- DP16 / local DP8 / TP1
- `deepep_high_throughput`
- `CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7`
- `VLLM_DEEPEP_BUFFER_SIZE_MB=256`
- `max_tokens=1`, three measured requests

Latest validated run:

| job | warmup | request 0 | request 1 | request 2 | p50 | p95 |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `codex-pr989-physical-node-rank-timed-20260613T010841` | 0.423s | 0.474s | 0.458s | 0.464s | 0.464s | 0.474s |

Local checks:

- `python -m py_compile ep/bench/utils.py`
- `git diff --check`
- rebuilt `uccl.ep` in a temporary SkyRL-compatible build environment
- import check confirmed `get_physical_gpu_rank` is exported and the extra diagnostic `get_gpu_bdf` helper is not exported


ep/src/rdma.cpp
```diff
@@ -387,10 +388,20 @@ bool is_cuda_host_pointer(void* ptr) {
 #endif
 }
 
+int get_gpu_numa_node_from_bdf(std::string const& bdf) {
```
MaoZiming: Hi, for this, could you move this function to under util? https://github.com/uccl-project/uccl/blob/main/include/util/util.h#L1232?

MaoZiming:
Thank you for the PR! @eicherseiji

ep/src/rdma.cpp
```diff
@@ -399,15 +410,38 @@ void per_thread_rdma_init(ProxyCtx& S, void* gpu_buf, size_t bytes, int rank,
...
 
-  // Ranked by GPU idx
-  auto gpu_cards = uccl::get_gpu_cards();
+  // Map the physical GPU rank (gpu_idx) to its PCI BDF for NIC/NUMA affinity.
+  auto all_gpu_bdfs = uccl::enumerate_all_gpu_bdfs();
+  std::filesystem::path gpu_device_path;
+  std::string gpu_bdf;
+  if (gpu_idx >= 0 && gpu_idx < static_cast<int>(all_gpu_bdfs.size())) {
+    gpu_bdf = all_gpu_bdfs[gpu_idx];
+    gpu_device_path = std::filesystem::path("/sys/bus/pci/devices") / gpu_bdf;
```
MaoZiming, line 413: https://github.com/eicherseiji/uccl/blob/codex/ray-vllm-ht-proxy-device-split/ep/bench/utils.py#L601C13-L601C47 This is setting `nic_local_rank` to be `barrier_local_rank`, which is then setting `gpu_idx`, doesn't that means we are using `barrier_local_rank` to pick from `all_gpu_bufs`?

ep/src/rdma.cpp
eicherseiji: Done

ep/src/rdma.cpp
eicherseiji: Fixed. `nic_local_rank` is no longer derived from `barrier_local_rank`; it now comes from the CUDA device’s physical PCI BDF rank.

`barrier_local_rank` is only used for the local barrier slot, and `device_index` is the index into `CUDA_VISIBLE_DEVICES`

eicherseiji (commented):
Addressed comments; tested with vLLM Ray backend + GLM 4.7 across two 8xH100 nodes w/ EFA

MaoZiming (approved):
Thank you. This LGTM

PR #993, opened 2026-06-12
[EP] LL clean & Zero-token LL rank issues

shawlleyw:
# Summary

This PR fixes two remaining EP issues exposed by SGLang DeepEP serving:

- Synchronize low-latency buffer cleaning across ranks and keep zero-token ranks participating in LL dispatch/combine.
- Treat hosts with no visible verbs devices as pure NVLink/no-NIC deployments, avoiding RDMA host-allocation fallback.

Upstream already contains fixes for the original empty-token/topk host assertions. This PR covers the remaining low-latency desync and pure NVLink no-NIC behavior.

### 1. Low-Latency Clean Can Desynchronize Ranks

In SGLang with `--deepep-mode auto`, DP attention, and heterogeneous requests, some ranks can be idle/zero-token while others continue serving. On upstream baseline, the server can launch successfully and reach readiness, but later `/generate` requests hang.

Observed on upstream `34bdf4f4`:

- SGLang server reached readiness.
- No old `topk_idx_ptr != 0` host assert fired.
- First heterogeneous `/generate` timed out after 180s.
- Scheduler processes spun hot.
- No Python traceback or pointer assert was emitted.

Root cause: `clean_low_latency_buffer` zeroes each rank's signaling buffer without a cross-rank barrier. A faster rank can write an IPC sentinel into a peer's buffer while that peer is still cleaning. The peer then wipes the sentinel and later spins forever waiting for it.

### 2. Zero-Token LL Ranks Must Still Participate

Upstream's LL empty-token path avoided pointer asserts by returning early from `low_latency_dispatch` / `low_latency_combine`. That avoids the crash, but it can still break the LL protocol: non-empty ranks continue waiting for protocol participation from zero-token ranks.

During verification, after adding the clean barrier, this surfaced as:

- First request returned HTTP 200.
- Later request either hit `TypeError: 'NoneType' object is not callable` in SGLang's `return_recv_hook` path or hung if replaced with a no-op hook.

The fix is to allow zero-token ranks to pass null data pointers, but still launch the LL protocol phases.

### 3. No-NIC Environments Should Not Use RDMA Host Fallback

For the RDMA scratch buffer, upstream uses host-pinned memory when GPU memory cannot be registered for RDMA. That is correct when NICs exist but GPU Direct RDMA registration fails.

However, in pure NVLink/no-NIC environments, `ibv_get_device_list()` returns no devices, causing the registration probe to fail. Upstream then incorrectly treats this as "must use host allocation," even though no RDMA path exists.

### Low-Latency SGLang Hang

The repro was run with:

- UCCL upstream baseline `34bdf4f49c8e33d618b983afc34fc5bec8686cfa`
- UCCL PR branch `upstream-ep-fixes`
- UCCL patched HEAD `e1533b4298b209f6c46e98b38205fe2c90d81371`
- SGLang `0.5.14.dev131+g880e6f66f`
- SGLang checkout `127b9e3283f7c2a43234b852ff5c9f1796d53624`
- PyTorch CUDA build from the SGLang venv
- 8 GPUs on one node

Hardware / software environment:

- OS: Debian GNU/Linux 12
- GPU: 8x NVIDIA H200, full NVLink connection
- NVIDIA driver: `535.161.08`
- CUDA runtime reported by `nvidia-smi`: `13.2`
- CUDA toolkit / `nvcc`: `13.2` (`V13.2.78`)
- Python: `3.11.2`
- PyTorch: `2.11.0+cu130`
- PyTorch CUDA: `13.0`
- GCC/G++: `12.2.0`

Build/install EP from the UCCL checkout into the SGLang environment:

```bash
cd /path/to/uccl/ep
python setup.py clean install
```

Launch SGLang from the SGLang checkout. The `PYTHONPATH` is only needed when running directly from a source checkout:

```bash
cd /path/to/sglang

export PYTHONPATH=/path/to/sglang/python:${PYTHONPATH:-}
export NO_PROXY=127.0.0.1,localhost,::1
export no_proxy=127.0.0.1,localhost,::1
export NCCL_SOCKET_FAMILY=AF_INET
export CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7
export SGLANG_DEEPEP_NUM_MAX_DISPATCH_TOKENS_PER_RANK=512
export NVSHMEM_QP_DEPTH=2048

python -m sglang.launch_server \
  --model-path Qwen/Qwen3-235B-A22B-Instruct-2507-FP8 \
  --trust-remote-code \
  --load-format dummy \
  --quantization fp8 \
  --tp-size 8 \
  --ep-size 8 \
  --dp-size 8 \
  --enable-dp-attention \
  --enable-dp-lm-head \
  --moe-a2a-backend deepep \
  --deepep-mode auto \
  --disable-cuda-graph \
  --ep-num-redundant-experts 0 \
  --ep-dispatch-algorithm dynamic \
  --host 127.0.0.1 \
  --port 30002 \
  --dist-init-addr 127.0.0.1:63800
```

Wait for SGLang readiness, then send heterogeneous requests. A launch-only smoke test is not sufficient; upstream baseline reaches readiness and then hangs during serving.

```bash
for i in 1 2 3 4 5 6 7 8; do
  PROMPT=$(printf 'Hi %.0s' $(seq 1 $((i * 5))))
  curl --noproxy '*' -m 180 \
    -X POST http://127.0.0.1:30002/generate \
    -H 'Content-Type: application/json' \
    -d "{\"text\":\"$PROMPT\",\"sampling_params\":{\"max_new_tokens\":5,\"temperature\":0}}"
done
```

On upstream baseline, the server reaches readiness but serving can hang/time out.

To verify the CUDA graph path after applying the fix, remove `--disable-cuda-graph` and add a bounded graph batch size:

```bash
--cuda-graph-max-bs 8
```

### No-NIC Behavior

In a container or host where `/dev/infiniband` / verbs devices are not visible, run:

```python
import uccl.ep as ep

print(ep.can_register_rdma_gpu_buffer(0, 4096))
print(ep.rdma_buffer_should_use_host_alloc(0, 4096))
```

Expected after this PR:

```text
True
False
```

No-NIC should be treated as pure NVLink, not as failed GPU RDMA registration.

### Low-Latency Synchronization

- Extend `clean_low_latency_buffer` to accept `barrier_signal_ptrs`, `rank`, and `num_ranks`.
- Add `barrier_block<kNumRanks>` before and after zeroing the clean buffers.
- Pass existing `barrier_signal_ptrs_gpu`, `nvl_rank`, and `num_nvl_ranks` from `Buffer::clean_low_latency_buffer`.

This prevents a fast rank's IPC write from being wiped by a slower rank's clean.

### Zero-Token LL Participation

- Remove LL dispatch/combine early returns for zero-token ranks.
- Relax pointer assertions when `x_rows == 0` / `topk_rows == 0`.
- Still launch the LL protocol phases so all ranks remain synchronized.

### No-NIC Handling

- Add `has_any_nic()` using `ibv_get_device_list()`.
- Only use host-pinned RDMA fallback when a NIC exists and GPU memory registration fails.
- In no-NIC environments:
  - keep CUDA allocation for the RDMA scratch buffer,
  - report `can_register_rdma_gpu_buffer(...) == true`,
  - report `rdma_buffer_should_use_host_alloc(...) == false`.

## Test Plan

Verified on an upstream-based branch from `34bdf4f4`.

- Built and installed `uccl.ep`.
- Confirmed upstream baseline:
  - SGLang server reaches readiness.
  - Heterogeneous `/generate` request times out.
  - Schedulers spin hot.
  - No old empty topk pointer assert is emitted.
- Verified patched EP helper smoke:
  - `can_register_rdma_gpu_buffer(0, 4096) == True`
  - `rdma_buffer_should_use_host_alloc(0, 4096) == False`
- Verified SGLang Qwen3-235B dummy, `tp8 ep8 dp8`, DP attention, DeepEP auto:
  - `--disable-cuda-graph`: 8/8 heterogeneous requests returned HTTP 200.
  - `--cuda-graph-max-bs 8`: 8/8 heterogeneous requests returned HTTP 200.
- Ran `git diff --check`.


ep/src/uccl_ep.cc
```diff
@@ -1275,21 +1279,7 @@ class Buffer {
...
-      if (async) {
-        event = EventHandle(comm_stream);
-      } else {
-        stream_wait(compute_stream, comm_stream);
-      }
-      std::optional<std::function<void()>> hook;
-      return {event, hook};
-    }
-    EP_HOST_ASSERT(x_ptr != 0 && topk_idx_ptr != 0);
+    EP_HOST_ASSERT(x_rows == 0 || (x_ptr != 0 && topk_idx_ptr != 0));
```
MaoZiming, line 1311: Last time I tested in some zero-token settings, you will have `x_ptr` and `topk_idx_ptr` equals nullptr, so we need to early return to avoid crashing in the later code

shawlleyw: Yes, `x_ptr` / `topk_idx_ptr` can be null for zero-token ranks. The reason I avoided the early return is that LL dispatch still needs zero-token ranks to participate in the protocol.

In the SEND phase, each rank still writes per-expert count signals. For a zero-token rank, `num_tokens_sent == 0`, so it writes `-1`, which tells the destination rank "this source is done and sent zero tokens." Receivers spin while the slot is `0`, so skipping this can hang peers.

Also, a rank with `x_rows == 0` can still own local experts and receive tokens from other ranks, so it still needs the RECV phase to populate `packed_recv_count` / layout and copy incoming tokens.

The null `x_ptr` / `topk_idx_ptr` should be safe because they are only dereferenced in loops bounded by `num_tokens` or `num_tokens * num_topk`; those loops are skipped when `num_tokens == 0`.

MaoZiming (approved):
LGTM

PR #996, opened 2026-06-16
[P2P] Add Hygon DCU (DTK) support for P2P RDMA

Iwandy777:
## Description
  Adds support for Hygon DCU hardware to the P2P module via Hygon's DTK
  (HIP-compatible SDK). DTK differs from upstream ROCm in a few ABI details
  (missing DMA-BUF API and `hipStreamLegacy`, renamed pointer-attribute field,
  and a `hipMemoryTypeDevice` enum value mismatch when linked against torch's
  ROCm 6.1 runtime) — all handled via macros in `gpu_rt.h`.

  Adds two build targets: `Makefile.dtk` (DTK native runtime) and
  `Makefile.dtk-torch` (links torch's ROCm 6.1 runtime so p2p and torch share
  one HIP context, enabling zero-copy GPU tensor sharing).

  Also fixes two pre-existing test bugs unrelated to Hygon found along the way
  (hardcoded `Endpoint(4)`, wrong op-name casing, nonexistent `check_xfer_state`
  method in `test_ray_api.py`; nonexistent `remote_gpu_idx` kwarg in
  `test_engine_onesided_ipc.py`).

  Fixes # (no existing issue — opening this PR directly)

  ## Type of Change
  - [x] Bug fix
  - [x] New feature
  - [ ] Documentation update

  ## How Has This Been Tested
  Validated on two physical Hygon DCU servers over real RDMA hardware:
  - Single-node: RDMA-WRITE/READ, memory registration, compressed transfer,
    multi-process — all pass
  - **Cross-node**: RDMA-WRITE and RDMA-READ verified end-to-end between two
    separate machines (new `test_cross_node_rdma.py` / `_native.py`)
  - Known limitation: same-node cross-process HIP IPC doesn't work through
    torch's official ROCm 6.1 runtime on Hygon (DTK's own native IPC works
    fine) — a torch/ROCm runtime gap, not a UCCL code issue.

  - [x] Unit tests
  - [x] Integration tests
  - [x] Manual testing

  ## Checklist
  - [ ] I have run `format.sh` (clang-format-14 unavailable in this env)
  - [ ] I have run `build.sh` (no Hygon DTK target in its matrix; verified
    instead via direct build+test on real Hygon hardware)
  - [x] I have added tests.


include/util/gpu_rt.h
```diff
@@ -160,15 +167,35 @@ inline gpuError_t gpuMemGetAddressRange(void** base_ptr, size_t* size,
...
+#define GPU_DRIVER_GET_HANDLE_FOR_ADDRESS_RANGE_NAME \
+  "hipMemGetHandleForAddressRange"
+#else
+typedef int gpuMemRangeHandleType;
+#define GPU_MEM_RANGE_HANDLE_TYPE_DMA_BUF_FD 0
+#define GPU_DRIVER_GET_HANDLE_FOR_ADDRESS_RANGE_NAME ""
+#endif
 #define gpuPointerAttribute_t hipPointerAttribute_t
 #define gpuPointerGetAttributes hipPointerGetAttributes
+// The pointer-attribute memory type field was renamed in ROCm: DTK/HIP≤5.x
```
derekwin: Please make the comments more concise.

include/util/util.h
```diff
@@ -1196,7 +1196,7 @@ inline int get_dev_idx(void* ptr) {
   hipPointerAttribute_t attributes;
   hipError_t err = hipPointerGetAttributes(&attributes, ptr);
   if (err == hipSuccess) {
-    if (attributes.type == hipMemoryTypeDevice) {
+    if (gpuPointerMemoryType(attributes) == gpuMemoryTypeDevice) {
```
derekwin: Since we're already on the HIP branch, shouldn't the original code just work? Is this change specifically to address a runtime issue on the DTK platform?

p2p/efa_stub/include/infiniband/efadv.h
```diff
@@ -0,0 +1,30 @@
+// Minimal EFA stub header for platforms without the AWS EFA SDK (e.g., Hygon
```
derekwin: Perhaps we could use a build flag to conditionally compile the EFA code path on servers without EFA support, instead of using a stub.

p2p/rdma/rdma_context.cc
```diff
@@ -209,7 +209,7 @@ bool RdmaContext::is_gpu_pointer(void* ptr) {
     (void)gpuGetLastError();  // clear sticky error
     return false;
   }
-  return (attrs.type == gpuMemoryTypeDevice);
+  return (gpuPointerMemoryType(attrs) == gpuMemoryTypeDevice);
```
derekwin: Since we're already on the HIP branch, shouldn't the original code just work? Is this change specifically to address a runtime issue on the DTK platform?

p2p/tests/test_cross_node_rdma.py
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env python3
+"""
```
derekwin: For PyTorch-dependent tests, consider reusing existing test scripts (e.g., `test_engine_xxxx.py`). Also, while the native test scripts look great, I suggest making them more extensible—for instance, by adopting a pluggable design so they can easily run on other GPU platforms.

p2p/Makefile.rocm
```diff
@@ -43,7 +43,7 @@ CXXFLAGS := -O3 -shared -std=c++17 -fPIC \
 	-I$(HIP_INC) -I$(EFA_HOME)/include \
 	-I${CONDA_LIB_HOME}/../include \
 	$(DIETGPU_INC) $(CXXFLAGS_DIETGPU) \
-	-D__HIP_PLATFORM_AMD__ \
```
derekwin, line 46: Since DTK uses a separate Makefile for compilation, is this change for ROCm strictly necessary?

derekwin:
Thanks, great job! Just a quick reminder to run bash format.sh to format the code.

derekwin:
If possible, could you please share the configuration details and performance results?

include/util/util.h
Iwandy777: Good catch — reverted both to the original. On DTK 24.04.1
hipPointerAttribute_t still exposes .type/hipMemoryTypeDevice, so the
original code compiles and works unchanged. Dropped the macro entirely.

p2p/Makefile.rocm
Iwandy777: Not necessary — reverted (and Makefile.therock). Gated the DMA-BUF path
on #ifndef __UCCL_DTK__ in gpu_rt.h instead, so non-DTK builds stay
byte-identical to main.

p2p/efa_stub/include/infiniband/efadv.h
Iwandy777: The EFA provider sources include <infiniband/efadv.h> unconditionally for
type definitions, so a build flag would need #ifdefs across several shared
files. The self-contained stub touches no shared code and is referenced
only by Makefile.dtk's EFA_HOME. Happy to switch to a build flag if you'd
prefer.

p2p/tests/test_cross_node_rdma.py
Iwandy777: Removed the redundant torch-based test — the torch-free native variant
covers cross-node and is the primary path on DTK. Also made the native
tests more portable: the GPU runtime lib is now overridable via
UCCL_GPU_RT_LIB.

Iwandy777:
> Config: 2× Hygon DCU Z100L (gfx906), DTK 24.04.1, built with `make -f Makefile.dtk` (native DTK runtime / libgalaxyhip, torch-free). Verified: cross-process/cross-GPU IPC, same-node RDMA write/read, and cross-node RDMA write.
>
> Performance (UCCL P2P, same-node 2-GPU, GPU memory, fp16, 50 iters) — RDMA WRITE and READ are symmetric:
>
> | Size | WRITE | READ |
> |------|-------|------|
> | 1 KB | 1.30 Gbps / 6.3 µs | 1.16 Gbps / 7.1 µs |
> | 16 KB | 15.9 Gbps / 8.3 µs | 15.3 Gbps / 8.6 µs |
> | 64 KB | 27.0 Gbps / 19 µs | 26.4 Gbps / 20 µs |
> | 256 KB | 33.9 Gbps / 62 µs | 34.5 Gbps / 61 µs |
> | 1 MB | 38.1 Gbps / 220 µs | 37.6 Gbps / 223 µs |
>
> ~6–7 µs small-message latency, scaling to ~38 Gbps (4.7 GB/s) at 1 MB.

YangZhou1997:
@Iwandy777, thanks for the contribution! I think efadv.h is included inside libibverbs, so it should not be a blocker even on a non-efa platform

Iwandy777:
> @Iwandy777, thanks for the contribution! I think efadv.h is included inside libibverbs, so it should not be a blocker even on a non-efa platform

Thanks @YangZhou1997, you're right. Our servers were on Mellanox OFED, whose libibverbs omits the EFA provider, so efadv.h was missing. I've switched them to upstream rdma-core (which ships efadv.h, as you noted) and dropped efa_stub and the EFA_HOME include from Makefile.dtk — it now builds against the system header.

derekwin (approved):

derekwin:
LGTM! Thanks to @Iwandy777. 
Hi Yang (@YangZhou1997 ), I believe the code is ready to be merged.

YangZhou1997 (approved):
The GH200 fails not because of running job on that node. The PR gets compiled successfully

Issue #956, opened 2026-05-21
[Proposal] libfabric-CXI backend for UCCL-EP on HPE Slingshot


PanJason:
# [Proposal]: libfabric-CXI backend for UCCL-EP on HPE Slingshot

Hi UCCL team,

We're from the [Apertus](https://www.swiss-ai.org/) / Swiss AI training effort, working on large-scale open MoE training on [Alps](https://www.cscs.ch/computers/alps) (using HPE's Slingshot-11 Cassini NIC). For Apertus 2 we're evaluating replacements for the default NCCL all-gather-style MoE dispatcher, and UCCL-EP looks like a strong fit because Slingshot does not provide IBGDA and we can only use some CPU-proxy design for all2all communication.

On Alps the network interface uses libfabric's `cxi` provider, not typical ibverbs. Before we start prototyping, we'd like to ask:

1. **Does UCCL-EP currently have a libfabric transport path?** I saw EFA, Broadcom, and CX7 mentioned and supported. Is EFA implemented via libfabric (`efa` provider), and if so, is the transport layer abstracted enough that adding a `cxi` provider would be mostly provider substitution? 
2. **Is any CXI / Slingshot work already underway or planned?** If so, we'd be glad to test on our platform for debugging and help with performance benchmarking.
3. **If not, would a libfabric/CXI backend be welcomed?** Our intent is to keep the UCCL-EP API unchanged and map proxy operations to libfabric RMA, starting with low-latency mode. This might be useful because, as far as we know, many EU data centers actually use Slingshot because they were initially built for HPC, not exactly LLM training. If we can demonstrate performance gains, it would benefit many users.

## What we'd contribute

Assuming the answer to (3) is yes, we'd plan to upstream:

- A libfabric/CXI transport backend (low-latency mode first, then normal/high-throughput).
- Validation on real Apertus MoE training on Alps Slingshot
- Benchmarks vs. all-gather performances.

## What we plan to verify first currently

[...]

PR #997, opened 2026-06-16
Add CXI transport support for EP

PanJason:
## Description

This PR adds a libfabric/CXI transport path for UCCL EP so the EP proxy can run over HPE Slingshot/Cassini through the `cxi` provider.

Implementation details:

- Adds an `EpTransport` abstraction and a `CxiTransport` implementation, enabled with `USE_LIBFABRIC_CXI=1`.
- Initializes libfabric `FI_EP_RDM` endpoints on the CXI domain matching the local GPU rank (`cxi0`, `cxi1`, ...), requesting RMA, atomic, fence, HMEM, local, and remote communication capabilities.
- Uses per-peer CXI transport state in the proxy: `cxi_transports_by_rank_` is keyed by peer rank, each peer transport owns its local CXI endpoint/MRs, inserts the matching remote endpoint name into the libfabric AV, and `post_cxi_commands()` routes each WRITE/ATOMIC to the destination peer's transport. This is because` shared endpoint` decreased the performance significantly.
- Registers the EP main RDMA buffer and atomic buffer as CUDA HMEM memory regions with provider keys, and exchanges the CXI endpoint name plus MR keys through the existing proxy metadata path.
- Maps EP WRITE commands to `fi_write`: the proxy decodes the local and remote EP buffer offsets and posts a CXI RMA write from the registered local main buffer to the peer main-buffer MR/key.
- Maps EP atomic notifications/control commands to libfabric atomics: normal EP ATOMIC commands become fenced `FI_SUM` `fi_atomicmsg` operations into the peer atomic buffer; WRITE commands carrying an atomic signal post the payload `fi_write` first and then a fenced `FI_SUM` atomic notification so the completion associated with the WR id corresponds to the notification step.
- Uses a small host-registered operand pool for outgoing atomics and polls the CXI CQ to translate libfabric completions back into the proxy's existing WR acknowledgement flow.
- Adds generic build knobs for CXI builds (`USE_LIBFABRIC_CXI`, `LIBFABRIC_HOME`, `NUM_MAX_NVL_PEERS`) and relaxes the internode benchmark local-rank assumption for 4-GPU GH200 nodes.

Related issue: #956 

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Documentation update

## How Has This Been Tested?

- [ ] Unit tests
- [x] Integration tests
- [x] Manual testing

Formatting:

- Format checker was run.

Build/validation:

- Our Alps cluster uses Slurm scheduler, so we wrote customized sbatch scripts that installed `ep` inside the target container with `USE_LIBFABRIC_CXI=1` via `python3 setup.py install`.
- Full EP validation completed on Alps GH200 nodes with the CXI transport enabled.

| EP | Job | Result | FP8 dispatch | BF16 dispatch | Combine |
|---:|---:|---|---:|---:|---:|
| 8 | `2544644` | completed `0:0` | 40.15-40.37 GB/s RDMA (1495-1502 us) | 42.94-43.00 GB/s RDMA (2722-2724 us) | 42.20-42.33 GB/s RDMA (2765-2772 us) |
| 16 | `2544645` | completed `0:0` | 25.95-26.23 GB/s RDMA (4172-4203 us) | 25.92-26.31 GB/s RDMA (8064-8126 us) | 26.63-27.10 GB/s RDMA (7829-7928 us) |
| 32 | `2544646` | completed `0:0` | 22.84-23.05 GB/s RDMA (5234-5281 us) | 22.24-22.45 GB/s RDMA (10423-10519 us) | 22.12-22.50 GB/s RDMA (10399-10566 us) |

No error was seen in the log.

Note: the representative performance validation above used the same CXI implementation plus a local benchmark/process-affinity patch to pin torchrun forked processes by local rank. On GH200 this pinning is needed for representative numbers; without it, the clean-branch validation still completed but measured noticeably slower.

## Checklist

- [ ] I have run `build.sh` to verify compilation. The full `build.sh` container wheel build was not run; instead, the target Slurm/container validation built and installed `ep` with `USE_LIBFABRIC_CXI=1`.
- [ ] I have updated the documentation. No user-facing documentation was added for this transport backend in this PR.
- [ ] I have added tests. No standalone unit test was added; validation used the existing EP internode benchmark on real CXI hardware.


PanJason:
@MaoZiming Could you take a look and let us know what other tests are needed for merging?

YangZhou1997:
Hi @PanJason , thank you for the contribution! 

The compilation on non-CXI machines would trigger some errors, see https://github.com/uccl-project/uccl/actions/runs/27840541212/job/82398387232

PanJason:
Add the fix to conditionally include. Check again?

YangZhou1997:
Both tests passed, can you give a close look if possible? @MaoZiming

MaoZiming:
@PanJason Thank you! Taking a look

ep/include/uccl_ibgda.cuh
```diff
@@ -410,7 +409,6 @@ __forceinline__ __device__ void nvshmem_sync_with_same_gpu_idx(
       }
     }
 #endif
-    break;
```
MaoZiming, line 413: https://github.com/uccl-project/uccl/pull/998
Let's maybe continue discussion there. do you see any issues when you (1) remove the first break (2) toggle the second break on/off. was the " forwarder epoch/NVL receiver timeouts" only observed with first break removed + second break on, and disappear with first break removed + second break removed?

ep/src/proxy.cpp
```diff
@@ -190,6 +238,159 @@ void Proxy::set_bench_d2h_channel_addrs(std::vector<uintptr_t> const& addrs) {
...
+      c.remote_rkey = remote_infos_[peer].rkey;
+      c.remote_atomic_buffer_addr = remote_infos_[peer].atomic_buffer_addr;
+      c.remote_atomic_buffer_len = remote_infos_[peer].atomic_buffer_len;
+      c.remote_atomic_buffer_rkey = remote_infos_[peer].atomic_buffer_rkey;
+      CxiTransport* transport = cxi_transport_for_rank(peer);
+      if (!transport) {
+        throw std::runtime_error("CXI peer transport is not initialized");
+      }
+      transport->connect_peer(peer, remote_infos_[peer]);
+      if (proxy_trace_enabled()) {
```
MaoZiming: Let's clean the profiling code, e.g. proxy_trace_enabled, and the cmd_name, etc.

ep/src/proxy.cpp
```diff
@@ -190,6 +238,159 @@ void Proxy::set_bench_d2h_channel_addrs(std::vector<uintptr_t> const& addrs) {
...
+      transport->register_main_buffer(cfg_.gpu_buffer, cfg_.total_size,
+                                      cfg_.local_rank);
+      transport->register_atomic_buffer(atomic_buffer_ptr_, kAtomicBufferSize,
+                                        cfg_.local_rank);
+      local_infos_[peer] = transport->local_info();
+      if (!cxi_transport_) cxi_transport_ = transport.get();
+      cxi_transports_by_rank_[peer] = std::move(transport);
+    }
+
+    std::thread receiver_thread([this, num_ranks, my_rank]() {
```
MaoZiming: Let's try to reduce duplicated code, e.g. `std::thread receiver_thread` `send_connection_info_as_client`, etc. also appear on the non-cxi path

ep/src/proxy.cpp
```diff
@@ -110,6 +153,11 @@ double Proxy::avg_wr_latency_us() const {
 
 uint64_t Proxy::completed_wr() const { return completion_count_; }
 
+bool Proxy::use_cxi_transport() const {
+  char const* transport = std::getenv("UCCL_EP_TRANSPORT");
```
MaoZiming: We can cache whether the transport is cxi rather than getting from env every time

ep/src/proxy.cpp
```diff
@@ -1472,6 +1880,90 @@ void Proxy::barrier_check() {
...
+          transport->post_barrier_atomic_add(
+              /*dst_rank=*/0, kCxiInternalWrId, kArrivalSlot, 1);
+          ++cxi_outstanding_ops_;
+        }
+      }
+
+      if (cfg_.rank == 0) {
+        uint64_t const expected_arrivals =
+            seq * static_cast<uint64_t>(std::max(0, cfg_.num_nodes - 1));
+        if (load_cxi_barrier_word_sum(kArrivalSlot) < expected_arrivals) {
```
MaoZiming, line 1822: Nit: the barrier seq never wraps around.

ep/include/cxi_transport.hpp
```diff
@@ -0,0 +1,76 @@
...
+#include <rdma/fi_domain.h>
+#include <rdma/fi_endpoint.h>
+#include <rdma/fi_eq.h>
+#endif
+
+#include <list>
+#include <stdexcept>
+#include <vector>
+
+class CxiTransport final : public EpTransport {
```
MaoZiming, line 17: cc @YangZhou1997 I think eventually it would be nice to package all different kinds of transport (since we have more of them now), into a class, right now only cxi instantiates `EpTransport`.

ep/include/uccl_ibgda.cuh
PanJason: I will do some A/B testing to answer your question. One thing I wanna know:
- Shall I respond you here or in #998?
- Also the following commits should be included in #998 or here?

Reason why asking is because I feel this is a generic fix, rather than sth specific to CXI port.

MaoZiming: Thanks! Let's incldue that in #998 !

MaoZiming:
I also notice that the nproc_per_node=4 (< 8) doesn't work on other transport paths, e.g. on EFA. It's currently hard-coded 
https://github.com/uccl-project/uccl/blob/main/ep/include/common.hpp#L88
We also need to parameterize that for other transports if we want to relax nproc_per_node<8 in general. It might be good to keep that as a separate PR. 


https://github.com/uccl-project/uccl/blob/main/ep/src/rdma.cpp#L1453-L1454

ep/src/proxy.cpp
PanJason: Cleaned in [fbbaf3d](https://github.com/uccl-project/uccl/pull/997/commits/fbbaf3d29bb375e59e963b4d6395d4b735faba11)

ep/src/proxy.cpp
PanJason: These two are folded into `exchange_peer_connection_info` now. Also duplicated if clause for connection build is folded to `should_connect_peer`. Cleaned in [b0a4d68](https://github.com/uccl-project/uccl/pull/997/commits/b0a4d68512484a8c48f850550ab69d4a601a8f5f)

ep/src/proxy.cpp
PanJason: Cached in `use_cxi_transport_` var. See [f46f191](https://github.com/uccl-project/uccl/pull/997/commits/f46f191b2c4e7e45f1831c3ce5aa1d18a6abd8ba)

ep/src/proxy.cpp
PanJason: Remove the wrap for CXI code in [7227ea4](https://github.com/uccl-project/uccl/pull/997/commits/7227ea48676c7f377fc57ac55743beada15454b7)

PanJason:
> It might be good to keep that as a separate PR.

@MaoZiming  A separate PR that makes MAX_NUM_GPUS configurable is created here: 
https://github.com/uccl-project/uccl/pull/1005

PanJason:
@MaoZiming Anything else needed for merging?

MaoZiming (approved):
Thanks @PanJason ! I just tested it. It LGTM

PanJason:
@YangZhou1997 Could you please also take a look at your convenience?

PR #999, opened 2026-06-17
[P2P]: add CXI transport endpoint

fergusfinn:
Thanks for all your work on UCCL!

This PR adds a CXI/libfabric transport for the UCCL P2P endpoint used by NIXL. It is separate from the UCCL-EP CXI work in https://github.com/uccl-project/uccl/issues/956 and https://github.com/uccl-project/uccl/pull/997, but it targets the same Slingshot/Cassini `cxi` provider.

Would be good to know if there should be more code sharing between this and the earlier EP patch.

## Benchmarking/testing

Benchmarking was performed on the [Isambard](https://docs.isambard.ac.uk/specs/#system-specifications-isambard-ai-phase-1) cluster in the UK - 4xGH200 nodes, with HPE Slingshot and 4 x Cassini 25GBps NICs on each node. 

```bash
export UCCL_P2P_TRANSPORT=cxi
export UCCL_P2P_DISABLE_IPC=1
export UCCL_LIBFABRIC_SO=/opt/libfabric/lib/libfabric.so.1

# Run once per num_iovs in 1,2,4,8,16,32,64,128.
torchrun --nnodes=2 --nproc_per_node=1 --node-rank=<0|1> \
  --master_addr=127.0.0.1 --master_port=<port> \
  p2p/benchmarks/benchmark_uccl.py \
  --local-gpu-idx=<0|1> --device=gpu --sizes=1179648 --iters=16 \
  --num-iovs=<num_iovs> --mode=write --float-type=none
```

| num_iovs | total_bytes | GB/s |
| ---: | ---: | ---: |
| 1 | 1179648 | 19.60 |
| 2 | 2359296 | 21.58 |
| 4 | 4718592 | 22.56 |
| 8 | 9437184 | 23.34 |
| 16 | 18874368 | 23.81 |
| 32 | 37748736 | 24.00 |
| 64 | 75497472 | 24.10 |
| 128 | 150994944 | 24.17 |

Then, KVBench in NIXL (currently, kvbench doesn't have UCCL wired in, but it has a plan mode, and nixlbench does have uccl wired in):

```bash
 python3 main.py plan \
    --model ./examples/model_deepseek_r1.yaml \
    --model_config ./examples/block-tp1-pp8.yaml \
    --backend UCX \
    --source gpu \
    --destination gpu \
    --runtime_type ASIO \
    --num_iter 16 \
    --warmup_iter 16 \
    --num_threads 1 \
    --page_size 256 \
    --format json
    
  nixlbench \
    --runtime_type ASIO \
    --backend UCCL \
    --worker_type nixl \
    --initiator_seg_type VRAM \
    --target_seg_type VRAM \
    --scheme pairwise \
    --mode SG \
    --op_type WRITE \
    --total_buffer_size 351535104 \
    --num_initiator_dev 1 \
    --num_target_dev 1 \
    --start_block_size 1179648 \
    --max_block_size 1179648 \
    --start_batch_size 298 \
    --max_batch_size 298 \
    --num_iter 16 \
    --warmup_iter 16 \
    --large_blk_iter_ftr 16 \
    --num_threads 1
 ```
 
| access | page_size | block_size | batch_size | total_bytes | GB/s | avg_lat_us | avg_prep_us | avg_post_us | avg_tx_us |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| block | 256 | 1179648 | 298 | 351535104 | 23.141011 | 51.0 | 12.0 | 9.0 | 15165.0 |

Also wired it in to vLLM, and vLLM disaggregated prefill completes for Qwen3-0.6b with direct NIXL telemetry showing expected KV transfer volume.

### CODE REVIEW

File: p2p/Makefile [hunk 1]
```diff
@@ -1,19 +1,27 @@
...
-# dlopen/dlsym (see *_dl.h). No transport libraries are linked.
+# libibverbs, libnccl, libefa, and optional libfabric/CXI are loaded
+# dynamically at runtime via dlopen/dlsym (see *_dl.h).
 #
-# Runtime: UCCL_P2P_TRANSPORT=rdma|nccl|efa (default: rdma)
+# Runtime: UCCL_P2P_TRANSPORT=rdma|nccl|efa|cxi (default: rdma)
 
 # DietGPU optional integration
 USE_DIETGPU ?= $(shell echo $${USE_DIETGPU:-0})
+USE_CXI ?= $(shell echo $${USE_CXI:-0})
```
praveingk: Could we drop the `USE_CXI=1 ` build flag entirely and follow the same pattern we already use for EFA in `rdma/efadv_dl.h`?

The runtime side of this PR is already fully dynamic — libfabric.so is dlopen'd, never linked. The only thing `USE_CXI` is actually gating is the compile-time include of <rdma/fabric.h> and friends. For EFA we sidestep that by not including `<infiniband/efadv.h> ` at all and instead locally redeclaring just the structs / enums / function-pointer types we use, then dlsym'ing at runtime. Same trick would work for libfabric here.

p2p/engine.cc
```diff
@@ -301,6 +327,14 @@ Endpoint::Endpoint() : local_gpu_idx_(INVALID_GPU), passive_accept_(false) {
 
   if (is_nccl_transport()) {
     ep_ = std::make_shared<NCCLEndpoint>(local_gpu_idx_, 0);
+#ifdef USE_CXI
```
praveingk: Same here, it might be confusing from usability PoV, that we have set `USE_CXI` during build, and also at runtime.

p2p/engine.cc
```diff
@@ -670,9 +704,14 @@ bool Endpoint::read(uint64_t conn_id, uint64_t mr_id, void* dst, size_t size,
   FifoItem curr_slot_item = slot_item;
   curr_slot_item.size = size;
 
-  while (uccl_read_async(ep_, conn, mhandle, dst, size, curr_slot_item,
-                         &ureq) == -1)
-    ;
+  int rc = UCCL_POST_TRANSIENT;
+  while (rc == UCCL_POST_TRANSIENT) {
```
praveingk: Looping on `UCCL_POST_TRANSIENT` in engine.cc is a problem for the existing backends.

Old loops retried on `-1` because that's what RDMA / NCCL / EFA return for transient queue-full (SQ full, ENOMEM on `ibv_post_send`). New loops retry only on -2, which only CXI produces — and the diff doesn't touch the other backends. So transient `-1` from RDMA / NCCL / EFA now hard-fails and returns false to the caller instead of retrying.

Two fixes:

1. Drop-in: while (rc == UCCL_POST_TRANSIENT only for CXI transport until backends are converted.
2. Proper: convert transient returns in rdma_endpoint.cc, nccl_endpoint.cc, and rdma/providers/* from -1 to UCCL_POST_TRANSIENT, reserve -1 for real errors.

praveingk (commented):
Thanks for this contribution, and it's great to see that you had validated with nixlbench. Yes, you are right that KVbench is not wired up yet.  I have a few comments which could be looked into.

p2p/Makefile
fergusfinn: sure! what's the pattern here - [rdma/efadv_dl.h](https://github.com/uccl-project/uccl/blob/main/p2p/rdma/efadv_dl.h) does seem to have #include <infiniband/efadv.h>. So the intended split is: require the headers for provider libs at compile time, but then be dynamic at runtime?

fergusfinn: latest commit does this, but i can put in a whole ABI shim if the goal is no compile time dep.

praveingk: Yes, just noticed `efadv_dl.h` is included, Sorry. Yes, we keep the headers for compile time, but linking is done dynamically at runtime.

praveingk:
@fergusfinn You may need to run `./format.sh` for the clang check to pass in CI

p2p/engine.cc
fergusfinn: ok, went with 1 with a note re. converting transient returns. i can open a followup to rationalise the use of UCCL_POST_TRANSIENT across the backends if i get a chance

YangZhou1997:
Thank you @fergusfinn for the contribution. Seems some header missing compilation errors on machines without libfabric/CXI installed, see the above workflow run

fergusfinn:
hi @YangZhou1997 - from the previous review from @praveingk i thought that including the various transport headers at compile time was the standard (i.e. [rdma/efadv_dl.h](https://github.com/uccl-project/uccl/blob/main/p2p/rdma/efadv_dl.h) has `<infiniband/efadv.h>`). Is the idea to vendor them? alternatively, i could update the PR to include them in the CI runner

YangZhou1997:
I see. My bad. I think your current approach is right. Our CI servers just do not have libfabric installed, thus triggering errors. I have created a PR on top of yours: https://github.com/doublewordai/uccl/pull/14, and it should pass our CI server.

YangZhou1997:
@praveingk the CI run looks good! Hope this may help your reviewing process

praveingk (approved):
Thanks for addressing the comments @fergusfinn . Looks good to me.

fergusfinn:
Anything still required to merge @praveingk?

praveingk:
> Anything still required to merge @praveingk?

Apologies. Let me do it now.

PR #1014, opened 2026-07-03
p2p/cxi: fail closed on FI_EIO instead of retrying forever

fergusfinn:
## Problem

`CxiEndpoint::post_rma()` classified both `-FI_EAGAIN` and `-FI_EIO` as `UCCL_POST_TRANSIENT`. The engine's post-retry loop (`is_retryable_post_failure`) therefore spun on `FI_EIO` **indefinitely**. `FI_EIO` is a hard transport error, not a transient out-of-resources condition, so the retry never succeeded and never surfaced — no completion and no error ever reached NIXL, and KV transfers stalled forever in `WAITING_FOR_REMOTE_KVS`.

Observed as a decode-side hang in DeepSeek-V4 prefill/decode disaggregation over CXI: prefill healthy, decode stuck, no error line.

## Fix

- Only `-FI_EAGAIN` is treated as transient (drive CQ, yield, retry).
- Any other non-zero post result falls through to the failure path and logs `ERROR`.
- On `-FI_EIO` specifically, the CQ error queue is first drained (`drain_cq_errors`) for provider-level detail and the failure is logged `FATAL`. `UCCL_LOG(FATAL)` aborts the process — intentional: a fatal transport fault should fail-stop rather than continue with a half-broken KV path. The drain is confined to this about-to-abort path because it consumes CQ error entries that may belong to other inflight ops, which would otherwise never be marked failed by `poll_cq_locked()`.

Confined to `p2p/cxi/cxi_endpoint.cc`; no change needed in `engine.cc` (its `rc < 0` bail already fires once `post_rma` stops mislabelling `FI_EIO` as transient).

### DIFF
--- a/p2p/cxi/cxi_endpoint.cc
+++ b/p2p/cxi/cxi_endpoint.cc
@@ -105,6 +105,35 @@ std::string cq_error_string(fid_cq* cq, fi_cq_err_entry const& err) {
   return fi_strerror(err.err);
 }
 
+// Drain and stringify any pending completion-queue error entries. Used to
+// attach provider-level detail (prov_errno, op_context, ...) to a failed RMA
+// post so hard failures like FI_EIO are diagnosable rather than opaque.
+std::string drain_cq_errors(fid_cq* cq) {
+  if (!cq) return "";
+
+  std::ostringstream os;
+  bool wrote = false;
+  for (int i = 0; i < 16; ++i) {
+    fi_cq_err_entry err{};
+    ssize_t rc = fi_cq_readerr(cq, &err, 0);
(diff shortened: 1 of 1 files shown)


fergusfinn:
an aside - there are quite a few of these kind of teething pain PRs that come up from testing the CXI backend. I've discussed getting some hardware on our cluster for running CI if that would be helpful? 

Otherwise if there's any way of my making the review process easier let me know - i'd be happy to be a collaborator/codeowner on the CXI parts of the stack

praveingk:
Thanks @fergusfinn for fixes. Would be great to have CI on CXI.

praveingk (approved):

PR #1018, opened 2026-07-13
  [P2P] convert interval tree test to proper pytest with full coverage

bhavyam2:
## Description
Rewrites `p2p/tests/test_util_interval_tree.py` from a print-based demo into 22 pytest assertions covering all six `ClosedIntervalTree` methods.

Previously the test file had no `assert` statements — CI could not catch any regression in this class. Now covered: `add`, `remove`, `query_containing`, `query_overlap`, `query_exact_match`, and `clear`, including edge cases (empty tree, single-point intervals, remove on non-existent entry, `ValueError` on inverted interval).

No related issue.

## Type of Change
- Bug fix — the missing assertions meant regressions would go undetected, which is a bug in the test suite.

## How Has This Been Tested?
Include any tests here. 

- pytest p2p/tests/test_util_interval_tree.py
- 22 passed in 0.04s
- Requires only pytest + intervaltree. No GPU or RDMA hardware needed.
- Check Unit tests.

### DIFF
--- a/ep/bench/utils.py
+++ b/ep/bench/utils.py
@@ -23,15 +23,6 @@ except ImportError as exc:
     sys.stderr.write("Failed to import uccl.ep\n")
     raise
 
-# import deep_ep as ep
-try:
-    from uccl import ep
-except ImportError as exc:
-    import sys
-
-    sys.stderr.write("Failed to import uccl.ep\n")
-    raise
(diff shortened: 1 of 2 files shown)


YangZhou1997 (approved):
LGTM!

PR #1020, opened 2026-07-14
ep: size internode TMA sender buffer for hidden up to 8192

whn09:
## Problem

The normal internode **dispatch** and **combine** kernels stage each token through a per-sender-warp TMA shared-memory buffer of `kNumTMABytesPer(Sender)Warp` bytes and assert:

```cpp
num_bytes_per_token + sizeof(uint64_t) <= kNumTMABytesPer(Sender)Warp
```

where `num_bytes_per_token = align(hidden * sizeof(bf16) + sizeof(SourceMeta) + scales + topk, 16)`.

The buffer was a bare `16384`-byte constant. That fits `hidden=7168` (DeepSeek, ~14.4 KB/token) but **not `hidden=8192`**: the bf16 payload alone is `8192 * 2 = 16384 B`, so the token plus the 8-byte TMA barrier overflow the buffer and the kernel aborts:

```
Assertion failed: num_bytes_per_token + sizeof(uint64_t) <= kNumTMABytesPerSenderWarp
```

Note: dispatch may send fp8, but **combine always stages bf16**, so an fp8 dispatch does not avoid this — the sender buffer must hold a full bf16 token.

This blocks `hidden=8192` MoE models (e.g. LongCat-2.0) on the normal internode path.

## Change

Replace the two magic `16384` sender-warp constants with a single file-scope `constexpr` sized for `hidden` up to 8192, plus a `static_assert` tying it to `kMaxSupportedHidden`. The hidden-independent forwarder-warp buffer (`9248`) is untouched.

Shared memory stays within H200's 227 KB opt-in limit:

| kernel | smem |
|---|---|
| dispatch | `20480 * NUM_MAX_NVL_PEERS(8)` = 160 KB |
| combine  | `max(20480*8, 9248*kNumCombineForwarderWarps(16))` = 160 KB |

## Testing

The equivalent raise (the two sender-warp constants `16384 → 20480`) was compiled into our EP build (`TORCH_CUDA_ARCH_LIST=9.0+PTX`, sm90 / H200) and used with LongCat-2.0-FP8 (hidden=8192); with it the `kNumTMABytesPerSenderWarp` assertion no longer fires. This PR expresses the same 20480 value as a named `constexpr` + `static_assert` rather than two magic numbers. Behaviour for existing hidden sizes (e.g. 7168) is unchanged — the buffer only grows. A maintainer CI build is welcome to confirm the refactored form compiles identically.


YangZhou1997:
@MaoZiming , would be great if you could take a look to unblock @whn09

MaoZiming (approved):
LGTM. thanks for the contribution!

PR #1029, opened 2026-07-22
[CI] bring up gb10

YangZhou1997:
## Description

Fixes # (issue)

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [ ] Manual testing

## Checklist
- [ ] I have added tests.


ep/bench/buffer.py
```diff
@@ -116,7 +116,7 @@ def __init__(
             # cudaMalloc/cudaMallocHost-style allocation instead of a possibly
             # segmented caching-allocator mapping.
             scratch_dlpack, rdma_buffer_is_host_allocated = ep.get_rdma_buffer(
-                num_rdma_bytes, device_index, bool(low_latency_mode)
+                num_rdma_bytes, device_index
```
MaoZiming, line 119: Why do we need to modify this?

YangZhou1997: GB10 requires allocating memory from host. This `bool(low_latency_mode)` will enforce the LL mode to allocate from device.

MaoZiming: do you mean for the GB10 we will also have the HT test? That's why we are removing this? Maybe we can also clean the function signature of `get_rdma_buffer` in uccl_ep.cc if needed

YangZhou1997: No, GB10 only has the LL test. bool(low_latency_mode) forces allocating on device, while GB10 only has host memory

MaoZiming (approved):

PR #1030, opened 2026-07-22
[EP] Add UCCL_EP_DISABLE_CPU_AFFINITY to opt out of proxy thread CPU pinning

aqni:
## Description
uccl-ep pins its proxy threads to specific CPU cores by default. When uccl-ep is embedded in a serving framework that already performs its own NUMA-aware, per-rank CPU binding on the whole process before spawning uccl-ep's proxy threads, uccl-ep's pinning logic can fight with the framework's binding over the same cores, since both compute core indices independently.

Add `UCCL_EP_DISABLE_CPU_AFFINITY=1` to let proxy threads keep whatever affinity they inherited from their parent process, instead of uccl-ep re-pinning them. Disabled by default to preserve existing behavior.

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [ ] Unit tests
- [ ] Integration tests
- [x] Manual testing

## Checklist
- [ ] I have added tests.


YangZhou1997 (approved):
Thank you!

PR #1037, opened 2026-08-12
p2p: variable-length notification messages

### LINKED ISSUE #1015: [P2P] 256-byte notification buffer silently drops large NIXL notifications
`notify_msg_t` caps notification payloads at a hardcoded 256 bytes (`p2p/uccl_engine.h`, `p2p/util/common.h`). The NIXL UCCL plugin drops anything larger with only a sender-side log line — so the peer never learns a notification was lost.

NIXL clients assume notifications are arbitrary-size (the API documents no limit; UCX delivers any size). 

vLLM's KV-lease heartbeat (vllm-project/vllm#38027) batches all active request IDs into one notification, crossing 256 bytes at ~5 concurrent requests. 

Observed with vLLM PD disaggregation over NIXL+UCCL:

``` 
[uccl_backend.cpp] Notification message too large: 305 bytes, max: 256 bytes 
[worker.py] Releasing expired KV blocks ... retrieved by 0 remote worker(s) before lease expired 
```

Dropped heartbeats silently stop lease renewal; the producer releases KV blocks and transfers fail minutes later with no attributable error. SGLang's disagg NIXL backend and TensorRT-LLM's NIXL transfer agent coordinate over the same channel, so it looks like they'd hit the same issue, but I didn't test.

Notifications ride the OOB TCP channel: is there a reason these need to be fixed size? I think it would require changes in the NIXL plugin as well

fergusfinn:
`notify_msg_t` currently caps `name` and `msg` at 256 bytes each, and the NIXL plugin drops larger notifications with only a sender-side log line. NIXL clients treat notifications as arbitrary-size, so batched payloads (e.g. vLLM's KV-lease heartbeat) are silently lost and transfers fail minutes later with no attributable error.

This change makes both fields `std::string` and carries the payload over the existing explicitly-sized OOB channels in the RDMA, NCCL, and CXI endpoints, removing the cap and the silent drop.

Fixes #1015


derekwin (approved):
LGTM!

praveingk (approved):
Looks good to me. Thanks @fergusfinn

praveingk:
It's an unrelated error.

PR #1040, opened 2026-08-22
[p2p] lkey bug && make MR key accessors explicit

derekwin:
## Description
Local SGE entries must reference the memory region's lkey, but RDMASendRequest::get_local_key() was returning the rkey. This only worked on hardware where the lkey and rkey happen to match; on stricter implementations it would surface as a local protection error. Fix the SGE path to use the lkey, and split the ambiguous get_key_by_* accessors into explicit get_lkey_by_* / get_rkey_by_* pairs (with RemoteMemInfo renamed to get_rkey_by_*) so the two keys can't be mixed up again.

Fixes # (issue)

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
Include any tests here. 
- [x] Unit tests
- [x] Integration tests
- [x] Manual testing

## Checklist
- [ ] I have added tests.

### DIFF
--- a/p2p/endpoint_wrapper.h
+++ b/p2p/endpoint_wrapper.h
@@ -421,7 +421,7 @@ inline int prepare_fifo_metadata(GenericEndpoint const& ep, P2PMhandle* mhandle,
           serialize_fifo_item(remote_mem_info, out_buf);
           return 0;
         } else {
-          FifoItem remote_mem_info;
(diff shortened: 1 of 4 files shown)


YangZhou1997 (approved):
Nice catch!

PR #1043, opened 2026-08-28
[EP] Accept vLLM enable_shrink compatibility argument

### LINKED ISSUE #1042: [ep] DeepEP compatibility wrapper rejects vLLM 0.28 enable_shrink argument
**Problem Statement**

vLLM 0.28 always passes the enable_shrink keyword when constructing deep_ep.Buffer, even when expert-parallel fault tolerance is disabled

```
deep_ep.Buffer(
    group=group,
    num_nvl_bytes=num_nvl_bytes,
    num_rdma_bytes=num_rdma_bytes,
    low_latency_mode=True,
    num_qps_per_rank=num_qps_per_rank,
    allow_nvlink_for_low_latency_mode=True,
    allow_mnnvl=False,
    explicitly_destroy=True,
    enable_shrink=False,
)
```
The UCCL DeepEP compatibility wrapper does not accept `enable_shrink` in `ep/bench/buffer.py`. Multi-rank expert parallel initialization therefore fails with: 

`TypeError: Buffer.__init__() got an unexpected keyword argument 'enable_shrink'`

Reproduction

Start vLLM with:

```
vllm serve Qwen/Qwen3-30B-A3B-FP8 \
  --enable-expert-parallel \
  --all2all-backend deepep_low_latency \
  --data-parallel-size 2 \
  --data-parallel-size-local 2 \
  --tensor-parallel-size 1
```
Initialization reaches DeepEPLLAll2AllManager.get_handle() and then fails while constructing deep_ep.Buffer.

**Possible Fix**

Extend the compatibility wrapper constructor:

```
def __init__(
    self,
    group: dist.ProcessGroup,
    num_nvl_bytes: int = 0,
    num_rdma_bytes: int = 0,
    low_latency_mode: bool = False,
    num_qps_per_rank: int = 24,
    allow_nvlink_for_low_latency_mode: bool = True,
    allow_mnnvl: bool = False,
    explicitly_destroy: bool = False,
    enable_shrink: bool = False,
    is_intranode: Optional[bool] = None,
) -> None:
    if enable_shrink:
        raise NotImplementedError(
            "UCCL EP does not currently support enable_shrink=True"
        )
```

This preserves compatibility with normal vLLM operation, where `enable_shrink=False`, while clearly rejecting fault-tolerant shrinking until UCCL implements it.

Expected Behavior

- enable_shrink=False: initialize UCCL EP normally.
- enable_shrink=True: either implement shrinking or report a clear unsupported-feature error.

ZenAlexa:
## Description

vLLM v0.28 always passes `enable_shrink` when it constructs a `deep_ep.Buffer`, including the ordinary `False` case. Accept that keyword at the same constructor position as DeepEP so the UCCL wrapper remains drop-in compatible. `False` follows the existing initialization path; `True` fails before any CUDA query or resource allocation because rank shrinking is not implemented by UCCL EP yet.

Fixes #1042

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?
- [x] Unit tests: `uv run --isolated --no-project --python 3.10 --with pytest --with black pytest -q ep/deep_ep_wrapper/tests/test_buffer_compatibility.py`
- [ ] Integration tests
- [ ] Manual testing

The regression test imports the public `deep_ep.Buffer` wrapper with CPU-only stubs, verifies `enable_shrink=True` is rejected before CUDA initialization, and verifies `enable_shrink=False` reaches the unchanged initialization path. It restores imported `deep_ep` modules afterward so the stubs cannot affect later tests in the same process.

## Checklist
- [x] I have updated the constructor documentation.
- [x] I have added tests.

Additional checks: Black left both changed Python files unchanged; Python 3.10 `compileall`, same-process module-isolation verification, and `git diff --check` passed.

### DIFF
--- a/ep/bench/buffer.py
+++ b/ep/bench/buffer.py
@@ -83,6 +83,7 @@ class Buffer:
         allow_nvlink_for_low_latency_mode: bool = True,
         allow_mnnvl: bool = False,
         explicitly_destroy: bool = False,
+        enable_shrink: bool = False,
         is_intranode: Optional[bool] = None,
     ) -> None:
         """
@@ -103,9 +104,15 @@ class Buffer:
             explicitly_destroy: If this flag is set to True, you need to explicitly call `destroy()` to release resources;
                 otherwise, the resources will be released by the destructor.
                 Note: Releasing resources in the destructor may cause Python's exception handling process to hang.
+            enable_shrink: whether to enable shrink mode. UCCL EP does not currently support shrink mode.
             is_intranode: whether to force intranode-only proxy mode. If set to `None`, infer it from the
                 process-group topology automatically. Explicit `True` is rejected when the group spans multiple nodes.
         """
+        if enable_shrink:
+            raise NotImplementedError(
+                "UCCL EP does not currently support enable_shrink=True"
+            )
+
         if "LOCAL_RANK" in os.environ:
             device_index = int(os.environ["LOCAL_RANK"])
         else:
--- a/ep/deep_ep_wrapper/tests/test_buffer_compatibility.py
+++ b/ep/deep_ep_wrapper/tests/test_buffer_compatibility.py
@@ -6,6 +6,25 @@ from types import ModuleType
 import pytest
 
 
+@pytest.fixture
+def isolated_deep_ep_import(monkeypatch):
+    def is_deep_ep_module(name):
+        return name == "deep_ep" or name.startswith("deep_ep.")
+
+    previous_modules = {
+        name: module for name, module in sys.modules.items() if is_deep_ep_module(name)
+    }
+    for name in previous_modules:
+        sys.modules.pop(name)
+
+    yield monkeypatch
+
+    for name in list(sys.modules):
+        if is_deep_ep_module(name):
+            sys.modules.pop(name)
+    sys.modules.update(previous_modules)
+
+
 class _StubTorch(ModuleType):
     def __getattr__(self, name):
         placeholder = type(name, (), {})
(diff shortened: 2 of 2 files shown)


ep/bench/buffer.py
```diff
@@ -103,9 +104,15 @@ def __init__(
             explicitly_destroy: If this flag is set to True, you need to explicitly call `destroy()` to release resources;
                 otherwise, the resources will be released by the destructor.
                 Note: Releasing resources in the destructor may cause Python's exception handling process to hang.
+            enable_shrink: whether to enable shrink mode. UCCL EP does not currently support shrink mode.
             is_intranode: whether to force intranode-only proxy mode. If set to `None`, infer it from the
                 process-group topology automatically. Explicit `True` is rejected when the group spans multiple nodes.
         """
+        if enable_shrink:
```
MaoZiming, line 111: Do you want to take a look at how to enable ``enable_shrink?" Is there anything incompatible with UCCL-EP's design?

ZenAlexa:
Yes. I traced DeepEP's shrink implementation in deepseek-ai/DeepEP#370 against UCCL-EP: the rank-mask contract fits the low-latency kernels, but complete support also needs timeout-and-skip behavior in dispatch/combine/clean and a shrink-aware cleanup barrier. UCCL's CPU proxy adds another fault domain because non-success RDMA CQEs currently abort and the GPU D2H/QUIET waits have no cancellation path, so I would keep this PR scoped to the `False` compatibility regression and take `True` as a separate follow-up starting with proxy error isolation and the mask query/update/clean contract.

MaoZiming:
@ZenAlexa Thank you. That makes sense

MaoZiming (approved):

PR #1047, opened 2026-09-09
[EP] Optimize dispatch layout statistics for prefill

ZenAlexa:
## Description

I added a token-partitioned histogram for high-throughput EP routing statistics. Each thread reads one token's top-k entries, counts expert occurrences, and deduplicates destination ranks with a two-word bitset. Blocks accumulate into the existing output arrays after a count-clearing launch on the same stream.

The histogram handles at least 4,096 tokens, top-k of at least 8, 256–1,024 experts, and 8–128 ranks, with evenly partitioned experts and eight-peer RDMA groups. The existing scan handles the remaining shapes and uses warp-cooperative final reductions. Output layout, masked slots, and per-token rank membership retain their existing semantics.

## Type of Change

- [x] Performance optimization

## How Has This Been Tested?

I compiled the branch's `layout.cu` and `main@79b64ae` with CUDA 12.4.131 and ran both on an H100 80 GB HBM3. The native test checks all four outputs against a CPU reference, covering masked slots, repeated experts, hot routing, tail blocks, empty inputs, and dispatch boundaries. Correctness checks pass, including top-k 1 and 7 at the histogram threshold.

Graph-replay timings include the histogram's clearing launch. Each sample contains 100 calls, with repeated measurements in alternating implementation order. Inputs select eight distinct experts uniformly per token. The table reports median per-call latency.

| Tokens | Ranks | Experts | Main (µs) | This PR (µs) |
|---:|---:|---:|---:|---:|
| 4,096 | 8 | 256 | 35.898 | 4.695 |
| 4,096 | 16 | 256 | 38.194 | 5.333 |
| 16,384 | 16 | 256 | 141.056 | 5.609 |
| 4,097 | 128 | 1,024 | 77.900 | 49.377 |

The [upstream GH200 job](https://github.com/uccl-project/uccl/actions/runs/34392676258) builds this head and passes its two-node low-latency EP tests. High-throughput integration through `ep/bench/test_intranode.py` and `ep/bench/test_internode.py`, and AMD execution, remain pending.

## Checklist

- [x] Formatting and whitespace checks pass.
- [x] Native CUDA compilation passes.
- [x] Existing distributed benchmarks provide layout references and integration coverage.
- [ ] Run high-throughput intranode and internode integration tests.


ZenAlexa:
Could you run the GH200 and AMD EP benchmarks for this branch? I checked the native layout kernel on H100 and would like to cover its distributed consumers and HIP execution.

YangZhou1997:
Hi @ZenAlexa thank you for the pr! We unfortunately on long has access to AMD servers anymore lol

ZenAlexa:
Haha, got it 😄 I'll leave AMD execution marked as unverified. Thanks for running GH200 — I've updated the PR with what that job covers.

YangZhou1997:
What do you think? cc @MaoZiming

MaoZiming:
@ZenAlexa Thanks!
What does the Main (µs) test? Was it `get_dispatch_layout`?

ZenAlexa:
Yes, it's the native `get_dispatch_layout` launcher from `main@79b64ae`, timed with CUDA Graph replay on H100. I measured the routing-layout work, including the new path's count-clearing launch. The timings cover that stage; end-to-end dispatch and communication need the high-throughput integration run.

YangZhou1997 (approved):

PR #1052, opened 2026-09-17
[EP] Make the device barrier safe for concurrent rounds

### LINKED ISSUE #986: [EP] Normal mode + TBO: prefill deadlocks at `notify_dispatch` barrier
## Summary

When SGLang's two-batch overlap (TBO, `--enable-two-batch-overlap`) is
combined with UCCL DeepEP normal mode (`--moe-a2a-backend deepep
--deepep-mode normal`), every prefill iteration in which TBO actually
fires deadlocks. All ranks freeze inside `(intra|inter)node_prepare`
on the CPU spin-wait for `*moe_recv_counter ≥ 0`; the
`notify_dispatch` GPU kernel never finishes its barrier. GPU
utilization stays at 100 % while memory utilization sits at 0 % —
kernels spin without making progress.

Both intranode (`num_rdma_ranks=1`) and internode (`num_rdma_ranks>1`)
configurations exhibit the hang. UCCL DeepEP normal mode **without**
TBO runs to completion. Removing only `--enable-two-batch-overlap` is
the smallest single-flag flip between hang and pass. UCCL low-latency
mode is not affected.

## Environment

- 2 × AWS p5en (8 × H200 each) connected over EFA + libfabric.
- Model: `Qwen/Qwen3-235B-A22B-FP8`.
- UCCL: `uccl-project/uccl` PR #984
  (`29a71ac5d7521dcd532f622a93ca1e338d534ebd`, branch
  `ep-zero-token-residual-asserts`).
- SGLang: v0.5.9 (commit `54dbf47558872b2792e687067aadf7a1a45d91e5`)
  with two patches applied to the SGLang source tree:
  1. UCCL's `ep/bench/sglang/fix-tbo-empty-microbatch.patch` —
     `compute_split_seq_index` returns `None` instead of producing
     splits with empty microbatches.
  2. A 2-line `operations_strategy.py` filter so PP stages > 0 don't
     `AttributeError` on `is_layer_sparse` when reading `layers[0]`:
     ```python
     # sglang/srt/batch_overlap/operations_strategy.py — top of init_new_tbo
     from sglang.srt.layers.utils import PPMissingLayer
     layers = [layer for layer in layers if not isinstance(layer, PPMissingLayer)]
     ```

## Reproduction

Replace `$NODE0_IP` with node 0's VPC IP, `$VENV/bin/python` with the
SGLang venv, and write `$DEEPEP_CFG` to a path reachable from both
nodes:

```json
{
  "normal_dispatch": {
    "num_sms": 24,
    "num_max_nvl_chunked_send_tokens": 16,
    "num_max_nvl_chunked_recv_tokens": 512,
    "num_max_rdma_chunked_send_tokens": 16,
    "num_max_rdma_chunked_recv_tokens": 512
  },
  "normal_combine": {
    "num_sms": 24,
    "num_max_nvl_chunked_send_tokens": 16,
    "num_max_nvl_chunked_recv_tokens": 512,
    "num_max_rdma_chunked_send_tokens": 16,
    "num_max_rdma_chunked_recv_tokens": 512
  }
}
```

UCCL/EFA env on both nodes:

```bash
# venv on PATH so JIT-built kernels (rope/cos_sin_cache via
# tvm_ffi.cpp.extension) can find `ninja`.
export PATH=$VENV/bin:$PATH
# Local HF cache only; avoid a re-download stall on filelock.
export HF_HUB_CACHE=/path/to/hf/cache/hub HF_HUB_OFFLINE=1
export PYTHONPATH=/path/to/uccl/ep/deep_ep_wrapper:$PYTHONPATH
export LD_LIBRARY_PATH=/path/to/build_libfabric/lib:/opt/amazon/ofi-nccl/lib/x86_64-linux-gnu:/opt/nccl/build/lib:/usr/local/cuda/lib64:/opt/amazon/efa/lib:/opt/amazon/openmpi/lib:/opt/amazon/ofi-nccl/lib:$LD_LIBRARY_PATH
export NVSHMEM_REMOTE_TRANSPORT=libfabric NVSHMEM_LIBFABRIC_PROVIDER=efa
export FI_PROVIDER=efa FI_EFA_USE_DEVICE_RDMA=1
export UCCL_SOCKET_IFNAME=enp71s0 UCCL_IB_GID_INDEX=0 UCCL_EP_CPU_TIMEOUT_SECS=300
# deep_gemm 2.5.0 lacks `get_compile_mode`; the precompile hook would
# AttributeError on the first FP8 GEMM and kill the scheduler before the
# TBO hang.
export SGLANG_JIT_DEEPGEMM_PRECOMPILE=0
```

**ep16 (internode, `num_rdma_ranks=2`) — node 0:**

```bash
$VENV/bin/python -m sglang.launch_server \
  --model-path Qwen/Qwen3-235B-A22B-FP8 --trust-remote-code \
  --dtype bfloat16 --host 0.0.0.0 --port 30000 \
  --nnodes 2 --node-rank 0 --dist-init-addr $NODE0_IP:29500 \
  --tp-size 16 --ep-size 16 --dp-size 16 --pipeline-parallel-size 1 \
  --enable-dp-attention --enable-dp-lm-head --moe-dense-tp-size 1 \
  --attention-backend fa3 --skip-server-warmup \
  --moe-a2a-backend deepep --deepep-mode normal --deepep-config $DEEPEP_CFG \
  --enable-two-batch-overlap --tbo-token-distribution-threshold 0.48 \
  --mem-fraction-static 0.85 --max-running-requests 512 \
  --chunked-prefill-size 32768 --max-prefill-tokens 32768 \
  --disable-radix-cache --disable-cuda-graph --random-seed 42
```

Node 1 is the same command with `--node-rank 1`.

**ep8pp2 (intranode, `num_rdma_ranks=1`):** same on each node, swap
the parallelism block to:

```
--tp-size 8 --ep-size 8 --dp-size 8 --pipeline-parallel-size 2
```

Bench client (any host that can reach `$NODE0_IP:30000`):

```bash
$VENV/bin/python -m sglang.bench_serving \
  --backend sglang --base-url http://$NODE0_IP:30000 \
  --dataset-name random \
  --random-input-len 1024 --random-output-len 1 --random-range-ratio 1.0 \
  --num-prompts 1024 --max-concurrency 1024 \
  --request-rate inf --warmup-requests 0 --seed 42
```

The bench progress bar freezes within a few hundred reqs and never
advances. Confirm with py-spy on any scheduler:

```bash
sudo $VENV/bin/py-spy dump --pid <any sglang::scheduler_DP* pid>
# top frame:
#   internode_dispatch (deep_ep/buffer.py:1608)   for ep16
#   intranode_dispatch (deep_ep/buffer.py:1118)   for ep8pp2
# called from op_dispatch_b inside execute_overlapped_operations
```

`disable_cuda_graph=True` is set, so this is not a graph-capture
artifact.

> **Why `--random-input-len 1024` and not 2048.** SGLang's TBO splitter
> only fires when each rank has at least 2 sequences in its current
> chunked-prefill chunk. Under DP=16 the per-rank chunk is
> `chunked_prefill_size / dp_size = 32768 / 16 = 2048` tokens, which
> fits two 1024-token sequences but only one 2048-token sequence. With
> in=2048 the splitter returns `None`, TBO is silently disabled, and
> the hang doesn't appear (but neither does TBO).

## Failure signature

py-spy on every scheduler shows the same stack on every rank:

```
internode_dispatch (deep_ep/buffer.py:1608)
dispatch (deep_ep/buffer.py:971)
_dispatch_core (sglang/srt/layers/moe/token_dispatcher/deepep.py:458)
dispatch_b   (sglang/srt/layers/moe/token_dispatcher/deepep.py:410)
dispatch_b   (sglang/srt/layers/moe/token_dispatcher/deepep.py:801)
_execute     (sglang/srt/batch_overlap/two_batch_overlap.py:1055)
dispatch_b   (sglang/srt/batch_overlap/two_batch_overlap.py:1064)
op_dispatch_b (sglang/srt/models/qwen3_moe.py:378)
...
```

Under ep8pp2 (intranode) the GPU side additionally produces UCCL's
`barrier_block` timeout printout
(`uccl/ep/include/ep_utils.cuh:920-924`) on every rank: every other
rank reports `value=1024` (= `FINISHED_SUM_TAG`) on the slot of one
specific missing rank — the classic "rank N didn't reach the barrier"
signature.

## Root cause

SGLang's `MaybeTboDeepEPDispatcher` constructs **two** inner
`DeepEPDispatcher` objects when TBO is on, intending one per
microbatch:

```python
# sglang/srt/batch_overlap/two_batch_overlap.py:1041
class MaybeTboDeepEPDispatcher(BaseDispatcher):
    def __init__(self, **kwargs):
        num_inner_dispatchers = 2 if is_tbo_enabled() else 1
        if get_moe_a2a_backend().is_deepep():
            self._inners = [
                DeepEPDispatcher(**kwargs) for _ in range(num_inner_dispatchers)
            ]
```

But every `DeepEPDispatcher` falls back to a class-level singleton
`Buffer`:

```python
# sglang/srt/layers/moe/token_dispatcher/deepep.py
class DeepEPBuffer:
    _buffer = None
    @classmethod
    def get_deepep_buffer(cls, ...):
        if cls._buffer is not None:
            return cls._buffer
        ...
```

So both inner dispatchers drive the **same** `Buffer` instance. That
`Buffer` owns one `*moe_recv_counter` host page (allocated via
`cudaMallocHost` at `uccl/ep/src/uccl_ep.cc:493-498`), one
`comm_stream`, and one `barrier_signal_ptrs` array.

Inside `notify_dispatch`, `barrier_block`
(`uccl/ep/include/ep_utils.cuh:894-928`) uses paired
`atomicAdd_system` / `atomicSub_system` of `FINISHED_SUM_TAG = 1024`
on `barrier_signal_ptrs`. The protocol assumes that all `kNumRanks`
participants reach the barrier in the same logical "epoch": each
entry into `barrier_block` writes `+1024` on
`barrier_signal_ptrs[rank][t]` for every `t`, then waits for
`barrier_signal_ptrs[rank][*] <= 0`.

When microbatch A runs on the singleton `Buffer`, A's barrier signals
zero out across ranks. Microbatch B then reuses the same
`barrier_signal_ptrs` array. If two ranks reach the barrier with
different (A vs B) microbatch progress because of cross-rank
scheduling jitter, A's `atomicAdd / Sub` pairs mix with B's on the
same slot and the `<= 0` test either fires spuriously or never. We
observe "never" — the spin-wait at `uccl/ep/src/uccl_ep.cc:712-742`
(intranode) and the analogous block in `internode_prepare` hang
forever.

## Direct evidence for the shared-state race

We instrumented `barrier_block` (`uccl/ep/include/ep_utils.cuh:894`)
to log `barrier_signal_ptrs[rank][thread_id]` at entry, *before* the
`atomicAdd`/`atomicSub` pair. The protocol invariant is that this
slot must equal `0` at every barrier entry: every previous round's
`+1024` add and `-1024` sub from each rank cancel out. Rebuilt UCCL,
ran the bug-1 repro:

| Metric | Node 0 | Node 1 |
|---|---|---|
| Pre-entry violations (slot != 0) within ~60 s | **116,095** | **110,335** |
| Pre-entry value when violated | **all `-1024`** | **all `-1024`** |
| Eventual timeout values (existing instrumentation) | — | `+1024` on rank 0 thread 3 and rank 2 thread 3 |

The `-1024` pre-entry signature means the previous round's `atomicSub`
arrived *after* its paired `atomicAdd` from the same round was
supposed to have already canceled it — write reordering between
concurrent users of the same array. The protocol's `<= 0` spin-wait
accepts the corrupted entry, so most rounds finish despite the
corruption. Eventually a slot transiently shows `+1024` (the sub from
the previous round arrived before the add from the current round,
leaving an unbalanced add briefly visible) and the spin-wait, which
expects `<= 0`, hangs. This is the documented "rank N missing from
barrier" timeout.

This rules out alternative hypotheses (proxy command-queue race,
cross-rank fabric contention, missing call): the corruption is at
the level of the singleton `barrier_signal_ptrs` slots, written by
both microbatches' `notify_dispatch` kernels without serialization.

A confirming negative test: with one extra `torch.cuda.synchronize()`
inserted between every TBO stage in
`sglang/srt/batch_overlap/operations.py::execute_overlapped_operations`
(eliminating any GPU-side overlap between A and B), the same bench
runs to completion (1024/1024 reqs, ~25.5K tok/s vs ~512/1024 hang
in the unmodified case). This confirms the race is overlap-dependent
on shared `Buffer` state.

0z5a:
**Incremental review:** [Files changed](https://github.com/uccl-project/uccl/pull/1052/files) against `main`.

## Description

`barrier_block` used one balanced counter per (rank, peer) pair: an
`atomicAdd_system(+FINISHED_SUM_TAG)` immediately followed by
`atomicSub_system(-FINISHED_SUM_TAG)`, with `value <= 0` as the release test.
That protocol carries no notion of *which* barrier round a modification belongs
to, so it is only correct while the signal array is used by a single barrier
instance at a time. Two dispatch epochs that overlap reuse the same array; adds
and subs from different rounds then mix on the same cells and the spin-wait
never converges. This is the primitive that #986 reports hanging in
`notify_dispatch`.

This PR makes the protocol monotonic and self-describing:

* each rank installs one tag in **every** peer cell per round, including its own,
  so after `k` rounds every cell holds `k * kNumRanks` tags;
* a thread's own cell only ever receives other ranks' tags and therefore counts
  the rounds it has completed;
* releasing once the row minimum has caught up therefore requires every rank to
  have arrived at this round.

Nothing is ever subtracted, so a late addition for a future round can only push
a cell past an older round's target and can never make a cell look released
early. No extra state, allocation or host coordination is needed, and the signal
layout is unchanged: `uccl_ep.cc` already allocates `kNumRanks` cells per rank
(`barrier_signal_bytes == max_nvl_peers * sizeof(int)`), of which the previous
implementation only used `(rank, rank)` and `(peer, rank)`.

The row minimum is reduced with a warp shuffle (`kNumRanks` never exceeds a warp
for the supported topologies, asserted), so no shared memory and no extra
atomics are needed.

Fixes #986

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

## How Has This Been Tested?

New test `ep/tests/barrier_concurrent.cu` includes the real
`ep/include/ep_utils.cuh` and calls the shipped `barrier_block`, so it exercises
the code under review rather than a copy of it. It runs two epochs concurrently
over one signal array from separate processes (one per GPU, signal blocks shared
over CUDA IPC), like two normal-mode epochs sharing a `Buffer`, and also checks
that no rank exits a round while a peer is still behind it.

```
cd ep/tests
make barrier_concurrent CUDAFLAGS=-arch=sm_89
CUDA_VISIBLE_DEVICES=0 ./barrier_concurrent 0 4 400 2 rr1 30 &
CUDA_VISIBLE_DEVICES=1 ./barrier_concurrent 1 4 400 2 rr1 30 &
CUDA_VISIBLE_DEVICES=2 ./barrier_concurrent 2 4 400 2 rr1 30 &
CUDA_VISIBLE_DEVICES=3 ./barrier_concurrent 3 4 400 2 rr1 30 &
```

Results on 4x L20 (PCIe, no NVLink, 2 concurrent epochs, 400 rounds = 3200
barrier calls):

| protocol | calls | completed | premature releases |
|---|---|---|---|
| before | 3200 | **0** (every call times out) | 0 |
| after | 3200 | 3200 | 0 |

The same test passes at 2 and 3 ranks (1600 and 2400 calls), and the L20 build
of `uccl.ep` succeeds with `DISABLE_SM90_FEATURES=1 TORCH_CUDA_ARCH_LIST=8.9`.

- [x] Unit tests
- [ ] Integration tests
- [x] Manual testing

## Checklist

- [x] I have added tests.

## Scope and limitations

* This fixes the barrier primitive, which is the shared-state hazard #986
  reports. Concurrent epochs also share the receive buffers, `send_head` and the
  mapped `moe_recv_counter`, so overlapping normal-mode epochs still need
  per-epoch ownership of that state; this PR does not claim to close the whole
  issue on its own.
* The original report is 2x p5en (8x H200, EFA) with SGLang two-batch overlap.
  That environment was not available here, so the end-to-end TBO run is **not**
  reproduced. On the L20 host the unmodified `ep/bench/test_intranode.py` is
  already flaky at 4 ranks (it fails intermittently with
  `DeepEP timeout check failed` and `DeepEP timeout for dispatch receivers`
  before any change), so in-library end-to-end runs there are not a reliable
  signal either way; the controlled A/B above is the evidence for this change.
* `make barrier_concurrent CUDAFLAGS=-arch=sm_90` was not run (no Hopper device
  available). The change adds no architecture-specific instruction.


ep/include/ep_utils.cuh
```diff
@@ -914,30 +914,41 @@ __forceinline__ __device__ void barrier_block(int** barrier_signal_ptrs,
...
 
-  // Add self-ranks, sub other ranks
+  EP_DEVICE_ASSERT(kNumRanks <= blockDim.x);
+  EP_DEVICE_ASSERT(kNumRanks <= WARP_SIZE);
+
+  // One tag in each cell, including our own. Nothing is ever subtracted, so a
+  // late addition for a future round can only push a cell past an older round
+  // target and can never make a cell look released early.
   if (thread_id < kNumRanks) {
     atomicAdd_system(barrier_signal_ptrs[rank] + thread_id, FINISHED_SUM_TAG);
```
YangZhou1997, line 924: [P1] Preserve cross-rank arrival signaling

The remaining atomic operation only updates `barrier_signal_ptrs[rank]`, and the wait loop reads that same local row. Removing the write to `barrier_signal_ptrs[thread_id] + rank` leaves no peer-arrival signal. For example, with eight ranks and zeroed counters, rank 0 alone increments its eight cells to 1024; `min_value >= own` then succeeds even if none of the other seven ranks has entered the barrier. `notify_dispatch` can consequently consume unfinished peer metadata. Please restore a protocol whose release condition depends on every participating rank arriving.

ep/include/ep_utils.cuh
```diff
@@ -914,30 +914,41 @@ __forceinline__ __device__ void barrier_block(int** barrier_signal_ptrs,
...
           "DeepEP timeout check failed: rank = %d, thread = %d, value = "
           "%d)\n",
-          rank, thread_id, value);
+          rank, thread_id, own);
       trap();
     }
+    __syncwarp();
   }
-  __syncthreads();
 }
```
YangZhou1997, line 952: [P1] Restore block-wide synchronization before returning

The final `__syncthreads()` has been removed. For every warp after the first, all threads have `thread_id >= kNumRanks`, so `own` is 0 and `row` is INT_MAX; those warps immediately satisfy the release condition and return, even while the first warp is still waiting. Existing callers use 128-thread blocks (`notify_dispatch`) and 256-thread blocks (`clean_low_latency_buffer`), so these threads can read or clear buffers prematurely. The new `__syncwarp()` cannot synchronize different warps. Please restore the final block barrier independently of repairing the cross-rank protocol.

ep/include/ep_utils.cuh
```diff
@@ -914,30 +914,41 @@ __forceinline__ __device__ void barrier_block(int** barrier_signal_ptrs,
...
-                     : 0;
-    if (__all_sync(WARP_MASK, value <= 0)) break;
+    auto const own = thread_id < kNumRanks
+                         ? ld_volatile_global(barrier_signal_ptrs[rank] + thread_id)
+                         : 0;
+    auto const row = thread_id < kNumRanks
+                         ? ld_volatile_global(barrier_signal_ptrs[rank] + thread_id)
+                         : 0x7fffffff;
+    auto const min_value = __reduce_min_sync(WARP_MASK, row);
+    if (__all_sync(WARP_MASK, min_value >= own)) break;
```
YangZhou1997, line 941: [P1] Make the accumulating counter comparison safe across overflow

Each signed 32-bit cell now accumulates `FINISHED_SUM_TAG` (1024) without being decremented or reset. After 2,097,152 increments, it wraps to -2147483648. In the supported 2-8 rank layouts, the first warp also contains inactive lanes whose `own` value is 0, so the negative `min_value` makes their comparison fail. Even with all ranks caught up, the barrier cannot release and eventually traps on timeout. Please use wrap-safe generation tracking/comparisons and account for inactive lanes.

ep/tests/barrier_concurrent.cu
```diff
@@ -0,0 +1,217 @@
...
+__global__ void real_barrier_stress(int **ptrs, unsigned int *pre,
+                                    unsigned long long *fail,
+                                    unsigned long long *timeouts,
+                                    unsigned long long *calls, int nranks,
+                                    int iters, int use_real_barrier) {
+  int rank = (int)blockIdx.x;
+  int epoch = (int)blockIdx.y;
+  int tid = (int)threadIdx.x;
+  int **cells = ptrs + (size_t)(rank * MAX_EPOCHS + epoch) * nranks;
+  unsigned int const round = (unsigned int)(epoch + 1);
```
YangZhou1997, line 60: [P2] Track actual epoch and iteration progress in the regression test

`round` is computed once as `epoch + 1` and never advances with `it`. Every launch at lines 200-202 also uses `dim3(nranks)`, so `blockIdx.y` is always 0 and every stream uses epoch 0. Consequently all `pre` entries become 1 after their first arrival, and the test can report no premature releases when one rank reaches iteration 400 while another remains at iteration 1. Please pass a distinct epoch identity, advance the per-round arrival sequence, and deliberately delay a peer on later iterations to verify that neither a later round nor another epoch can satisfy the current barrier.

PR #1054, opened 2026-09-18
fix(p2p): make an accept aborted by stop_accept observable

0z5a:
**Incremental review:** [Files changed](https://github.com/uccl-project/uccl/pull/1054/files) against `main`.

## Problem

Destroying a UCCL-backed agent hangs forever when the endpoint never registered
memory — a teardown after a connection that was never used, or after memory
registration failed.

`Endpoint::accept()` waits for `engine_initialized_`:

```cpp
while (!engine_initialized_) {
  std::this_thread::sleep_for(std::chrono::milliseconds(1));
}
```

`engine_initialized_` is only set by `initialize_engine()`, which runs from
`reg()`. An endpoint that never registers memory never leaves that loop, and
`~Endpoint()` joins the accept thread — so the process can never exit.

Thread-dump evidence: all 69 threads parked in `futex_wait`; the main thread in
`nixlUcclEngine::~nixlUcclEngine()` joining its listener; the joined listener
inside the `RDMAEndpoint::uccl_accept` loop reached through
`EpollClient::event_loop`.

## Fix

`stop_accept()` already existed but was not observable from that wait, so
shutdown had no way to break it. Three changes:

1. Each endpoint exposes `accept_stopped()` over its existing stop flag —
   `RDMAEndpoint` and `NCCLEndpoint` return it, `CxiEndpoint` returns `false`
   because it has no stop-accept.
2. `Endpoint::accept()` checks both its own shutdown flags and the transport flag every iteration, including before a lazy transport exists, and returns `false` once stopped.
3. `RDMAEndpoint::uccl_accept()` keeps `peer_id` at `UINT64_MAX` until all required channels are present, so an aborted
   accept returns the invalid `ConnID` callers already reject, instead of a
   well-formed id pointing at peer 0 that then waits forever for channel
   metadata.

**The wait itself has to stay.** `initialize_engine()` runs from `reg()`, so a
legitimate accept can arrive before memory registration and must be able to wait
for it. Deleting the wait instead of making it interruptible was measured at
**~9800 guard trips per run**, the peer's connect refused, and the data path
going from 5/5 to 0/5.

## Review regression checks

Updated in 32a1d0d. Native build passed on RTX 5090/CUDA 13.0 and L40S/CUDA 12.8.

| Regression | Before | After |
| --- | --- | --- |
| Partial handshake cancellation (compiled actual method) | Invalid-ID assertion fails | Pass |
| Empty cancellation and complete handshake | — | Both pass |
| Lazy endpoint passive accept, no registration, destruction | Reviewer-reported hang | Native subprocess passes with 20 s timeout |

The following measurements predate this review revision and are retained as earlier evidence.

## Test result

Verified on a 2x L20 host, UCCL backend, no RDMA devices present (NCCL
transport), against the real agent lifecycle rather than a stub.

| probe | result |
| --- | --- |
| `del agent`, no peer, **UCX** control backend | agent release **30.83 ms**, exit 0 |
| `del agent`, no peer, **UCCL**, unpatched | release **never returns**, exit 124 (deadline) |
| `del agent`, no peer, **UCCL**, patched | release returns, `Engine destroyed` |

Lifecycle matrix, paired initiator/target transfers:

| case | op | initiator / target | runs | result | bandwidth |
| --- | --- | --- | --- | --- | --- |
| B4-DRAMWRITE | WRITE | DRAM / DRAM | 5 | PASS 5/5 | 3.33-7.02 GB/s |
| B4-DRAMREAD | READ | DRAM / DRAM | 5 | PASS 5/5 | 2.74-5.92 GB/s |
| B4-LONG | READ | DRAM / DRAM | 3 | PASS 3/3 | 2.56-6.27 GB/s |
| B5-DELAY2 | READ | DRAM / DRAM | 5 | PASS 5/5 | 3.43-5.98 GB/s |
| B2-VRAMREAD | READ | DRAM / VRAM | 2 | `TEARDOWN_HANG` | - |
| B2-VRAMWRITE | WRITE | DRAM / VRAM | 2 | `TEARDOWN_HANG` | - |

19/19 DRAM-path runs exit normally on both ends, so the fix does not disturb the
teardown chain when a connection did exist. The two VRAM cases are **blocked by
the environment, not by this change**: the host has no `nvidia_peermem`, so
`ibv_reg_mr_iova2` on device memory fails with `Bad address` (errno 14) and the
transfer never starts (`BLOCKED_ENV`).

Post-change teardown runs reach `Engine destroyed` **5/5** — i.e. `~Endpoint()`
completes — and the paired-transfer A/B keeps `PROCESS_EXIT rc=0` on both arms,
so the happy path is unaffected.

Timeout samples are counted as failures throughout; nothing was excluded or
retried away.

## Scope

No behaviour change on the happy path. `accept_stopped()` is only consulted
while already waiting for engine init, and `CxiEndpoint` keeps its existing
semantics by reporting `false`.

## Known residual (not fixed here)

1 of 3 multi-process WRITE runs had both endpoints print `PROCESS_EXIT rc=0`
while teardown was still killed by a watchdog at 174 s. Both ends had already
exited, so this is not the idle-accept deadlock; a deterministic regression
(20/20) does not reproduce it. Recorded as flaky rather than folded into this
change.


0z5a:
Force-updated this branch (`a2a4ec6c` -> `a8c8a181`) with a simplified formulation
of the same fix.

The earlier revision dispatched through a `uccl_accept_stopped()` helper in
`endpoint_wrapper.h` with an `if constexpr` / `std::decay_t` branch to special-case
`CxiEndpoint`. Each endpoint now exposes `accept_stopped()` over its own stop flag
instead, so the visitor branch disappears: **6 files, +28/-2** versus 5 files,
+41/-2 before.

Semantics are unchanged — `CxiEndpoint` still reports `false` because it has no
stop-accept. The verification in the description was produced on this hardware
against the same behaviour, and the post-change teardown runs reach
`Engine destroyed` 5/5.

No reviews had been submitted, so the rewrite is a clean replace rather than a
follow-up commit.

p2p/engine.cc
```diff
@@ -558,8 +558,12 @@ bool Endpoint::accept(std::string& ip_addr, int& remote_gpu_idx,
   // For demo purposes, simulate accepted connection
   conn_id = next_conn_id_.fetch_add(1);
 
-  // Wait until engine is intialized to get the correct local_gpu_idx_
+  // initialize_engine() runs from reg(), so accept may legitimately arrive
+  // first. The wait must also observe stop_accept(): an endpoint that never
+  // registers memory would otherwise spin here forever while ~Endpoint()
+  // blocks in join().
   while (!engine_initialized_) {
+    if (uccl_accept_stopped(ep_)) return false;
```
YangZhou1997: **[P1] Observe the destructor's stop request in the initialization wait**

The built-in passive-accept lifecycle still deadlocks for a lazy endpoint that never calls `reg()`. `~Endpoint()` sets `stop_` and `passive_accept_stop_` before joining the accept thread, but this new check reads the transport's separate `stop_accept_` flag. That flag is only set from the later future-wait loop, which cannot be reached while `engine_initialized_` is false.

I reproduced this by compiling the current `accept()` / `start_passive_accept()` / destructor code with a fake transport: starting passive acceptance and then destroying the endpoint times out; explicitly calling `stop_accept()` first returns. Please check the endpoint shutdown flags here or propagate `stop_accept()` before joining, and cover the no-registration passive-accept lifecycle.

p2p/rdma/rdma_endpoint.cc
```diff
@@ -551,6 +554,13 @@ ConnID RDMAEndpoint::uccl_accept(std::string& remote_ip, int* remote_gpuidx) {
 
   // Create and return ConnID
   ConnID conn_id;
+  if (peer_id == UINT64_MAX) {
```
YangZhou1997, line 556: **[P2] Keep cancelled partial handshakes invalid**

`peer_id` is assigned from `accepted_meta_` before checking that all `kQpNumPerChannel + 1` channels exist. If `stop_accept()` arrives after control-channel metadata is available but while data channels are still incomplete, the loop exits with a real `peer_id` and bypasses this new invalid-return branch. `uccl_accept()` then returns `sock_fd=0`, and `Endpoint::accept()` treats the aborted handshake as a successful connection.

A test using the current `uccl_accept()` body with a fake channel group containing only the control channel reproduces `peer_id=7` / `sock_fd=0` after cancellation; empty-cancelled and fully established controls behave correctly. Please only mark the accept successful when the channel-count condition is satisfied (or track an explicit completion flag), and return an invalid `ConnID` otherwise.

p2p/engine.cc
0z5a: Fixed in 32a1d0d: the lazy initialization loop now observes both endpoint shutdown flags before the transport exists. The real native no-registration start/stop/destruction subprocess regression passes (20-second timeout).

p2p/rdma/rdma_endpoint.cc
0z5a: Fixed in 32a1d0d: peer_id remains invalid until every required channel is present. A compiled harness using the actual uccl_accept method passes empty cancellation, partial-handshake cancellation, and complete acceptance; the previous method fails the partial-cancellation assertion.

